Issue #7, opened 2020-10-30
set up a documentation site


johnynek:
have a doc site that publishes and looks similar to the cats documentation (logos etc...)

PR #30, opened 2020-11-03
Add initial mdoc support

regadas:
* `mdoc` compile is tied to `Compile / paradox`.
* `README.md` is now a symlink to `index.md`


johnynek:
Thanks! We would need to also run mdoc in the CI no? or is it automatically done by doc? Did you very the CI fails if you break some code?

regadas:
@johnynek I was trying to tie it to the compile task but it seems to be a little bit tricky... the easy way was to add `docs/mdoc` as one of the steps. CI should fail if code breaks.

We will need to limit the docs build to 2.12/2.13 though ... I'm seeing some issues with dotty.

regadas:
@johnynek added a separate step for docs

johnynek:
thanks @regadas ! I don't really know much about the sbt plugin we are using to generate the yml, but it seems like it is doing the right thing.

Thanks.

johnynek:
Related to #7

PR #37, opened 2020-11-04
First pass at improving the Scaladoc comments.

non:
I'm trying to work through and write more descriptive Scaladoc comments.

In particular, I've coined the term _arresting failure_ to contrast with _epsilon failure_. I think this makes a lot of the prose easier to follow.

What do you think? cc @johnynek @rossabaker

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -42,18 +42,20 @@ import cats.implicits._
  *
  *   - Arresting failure: The parser fails to extract a value but does
  *     consume one-or-more characters of input. The input offset will
- *     be moved forward by the number of characters consumed. The
- *     consumed characters will not be available to any additional
- *     parsers which may run.
+ *     be moved forward by the number of characters consumed and all
+ *     parsing will stop (unless a higher-level parser backtracks).
  *
  * Operations such as `x.orElse(y)` will only consider parser `y` if
  * `x` returns an epsilon failure; these methods cannot recover from
  * an arresting failure. Arresting failures can be "rewound" using
- * methods such as `backtrack` or `softProduct`, which converts
- * arresting failures into epsilon failures. Since rewinding is
- * expensive and tends to make error reporting more difficult it is
- * not the default behavior.
-  */
+ * methods such as `x.backtrack` (which converts arresting failures
+ * from `x` into epsilon failures), or `softProduct(x, y)` (which can
+ * rewind successful parses by `x` that are followed by epsilon
+ * failures for `y`).
+ *
+ * Rewinding tends to make error reporting more difficult and can lead
+ * to exponential parser behavior it is not the default behavior.
+ */
 sealed abstract class Parser[+A] {
 
   /**
@@ -75,12 +77,12 @@ sealed abstract class Parser[+A] {
   }
 
   /**
-   * Attempt to parse all of `str` into an `A` value.
+   * Attempt to parse all of the input `str` into an `A` value.
    *
    * This method will return a failure unless all of `str` is consumed
    * during parsing.
    *
-   * `p.parseAll(s)` is equivalent to `(p <* * Parser.end).parse(s).map(_._2)`.
+   * `p.parseAll(s)` is equivalent to `(p <* Parser.end).parse(s).map(_._2)`.
    */
   final def parseAll(str: String): Either[Parser.Error, A] = {
     val state = new Parser.Impl.State(str)
@@ -133,6 +135,9 @@ sealed abstract class Parser[+A] {
    * this parser will return the matched substring instead of any
    * value that the underlying parser would have returned. It will
    * still match exactly the same inputs as the original parser.
+   *
+   * This method is very efficient: similarly to `void`, we can avoid
+   * allocating results to return.
    */
   def string: Parser[String] =
     Parser.string(this)
@@ -141,8 +146,12 @@ sealed abstract class Parser[+A] {
    * If this parser fails to match, rewind the offset to the starting
    * point before moving on to other parser.
    *
-   * This method will mostly be used before calling methos such as
-   * `orElse`, `~`, or `flatMap` which involve a subsequent parser
+   * This method converts arresting failures into epsilon failures,
+   * which includes rewinding the offset to that used before parsing
+   * began.
+   *
+   * This method will most often be used before calling methods such
+   * as `orElse`, `~`, or `flatMap` which involve a subsequent parser
    * picking up where this one left off.
    */
   def backtrack: Parser[A] =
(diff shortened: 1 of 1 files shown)


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -26,21 +26,44 @@ import cats.data.{Chain, NonEmptyList}
...
+ *     consume one-or-more characters of input. The input offset will
+ *     be moved forward by the number of characters consumed. The
+ *     consumed characters will not be available to any additional
+ *     parsers which may run.
+ *
+ * Operations such as `x.orElse(y)` will only consider parser `y` if
+ * `x` returns an epsilon failure; these methods cannot recover from
+ * an arresting failure. Arresting failures can be "rewound" using
+ * methods such as `backtrack` or `softProduct`, which converts
+ * arresting failures into epsilon failures. Since rewinding is
```
johnynek, line 53: Rewinding is actually cheap, but backtracking can result in exponential complexity and tends to make errors harder to locate.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -26,21 +26,44 @@ import cats.data.{Chain, NonEmptyList}
...
+ *     moved forward by the number of characters consumed.
+ *
+ *   - Epsilon failure: The parser fails to extract a value without
+ *     consuming any characters of input. The input offset will not be
+ *     changed.
+ *
+ *   - Arresting failure: The parser fails to extract a value but does
+ *     consume one-or-more characters of input. The input offset will
+ *     be moved forward by the number of characters consumed. The
+ *     consumed characters will not be available to any additional
```
johnynek, line 46: The "will not be available" but isn't great wording IMO. After an arresting error, nothing more will be parsed, full stop. It will always be returned to the user.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -26,21 +26,44 @@ import cats.data.{Chain, NonEmptyList}
...
+ *     parsers which may run.
+ *
+ * Operations such as `x.orElse(y)` will only consider parser `y` if
+ * `x` returns an epsilon failure; these methods cannot recover from
+ * an arresting failure. Arresting failures can be "rewound" using
+ * methods such as `backtrack` or `softProduct`, which converts
```
johnynek, line 52: This isn't quite right about softProduct. SoftProduct makes a success in the first followed by an epsilon failure in the second into an epsilon failure of the whole thing. By contrast, a normal product the first may have consumed and thus you have an arresting failure even though the second item was an epsilon.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -51,9 +74,14 @@ sealed abstract class Parser[+A] {
...
-  /** Parse all or fail
-    * same as (this <* Parser.end).map(_._2)
-    */
+  /**
+   * Attempt to parse all of `str` into an `A` value.
+   *
+   * This method will return a failure unless all of `str` is consumed
+   * during parsing.
+   *
+   * `p.parseAll(s)` is equivalent to `(p <* * Parser.end).parse(s).map(_._2)`.
```
johnynek, line 83: There is an extra `*` before Parser.end.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -72,162 +100,315 @@ sealed abstract class Parser[+A] {
...
-    * a value of None, else Some for the parsed value
-    */
+  /**
+   * Convert epsilon failures into None values.
+   *
+   * Normally if a parser fails to consume any input it fails with an
+   * epsilon failure. The `?` method converts these failures into
+   * None values (and wraps other values in `Some(_)`).
+   *
+   * If the underlying parser failed with other errors, this parser
```
johnynek, line 110: Do you want to say "arresting error" here?

non:
(On reading it here I notice the prose is fairly clunky in some places. I plan to do another pass before I remove the "wip" title -- feel free to comment on any problems you notice but as far as typos or misspellings you can also wait until it's closer to ready.)

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -72,162 +100,315 @@ sealed abstract class Parser[+A] {
...
-  /** Discard the value and get the substring that this parser
-    * matches
-    */
+  /**
+   * Return the string matched by this parser.
+   *
+   * When parsing an input string that the underlying parser matches,
+   * this parser will return the matched substring instead of any
+   * value that the underlying parser would have returned. It will
+   * still match exactly the same inputs as the original parser.
```
johnynek, line 135: You might mention this is very efficient and the same as .void.string internally.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -72,162 +100,315 @@ sealed abstract class Parser[+A] {
...
+   */
   def string: Parser[String] =
     Parser.string(this)
 
-  /** If this fails, rewind the offset to the starting point
-    * so we can be used with oneOf
-    */
+  /**
+   * If this parser fails to match, rewind the offset to the starting
+   * point before moving on to other parser.
```
johnynek, line 142: I would say "convert an arresting failure into an epsilon failure" at some point.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -72,162 +100,315 @@ sealed abstract class Parser[+A] {
...
+   *
+   * This combinator returns a product of parsers. If this parser
+   * successfully produces an `A` value, the other parser is run on
+   * the remaining input to try to produce a `B` value.
+   *
+   * If either parser produces an error the result is an error.
+   * Otherwise both extracted values are combined into a tuple.
+   *
+   * Backtracking may be used on the left parser to allow the right
+   * one to pick up after an error.
```
johnynek, line 163: This isn't quite right. If A fails B can never run, the only question is will you have an epsilon or arresting failure. If A needs backtracking but B does not, you might do A.backtrack ~ B which is different from (A ~ B).backtrack

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -72,162 +100,315 @@ sealed abstract class Parser[+A] {
...
+   * failure, the parser will "rewind" to the point before `x` began.
     */
   def soft: Parser.Soft[A] =
     new Parser.Soft(this)
 
-  /** A parser that succeeds consuming nothing if this
-    * current parser would fail
-    */
+  /**
+   * Return a parses that succeeds (consuming nothing, and extracting
```
johnynek, line 243: S/parses/parser/

johnynek (commented):
Thanks. These comments are really nice!

I made a few comments.

johnynek (approved):
This is really great! Thanks for making the docs so much clearer.

PR #38, opened 2020-11-04
Test some distributivity laws

johnynek:
I had naively hoped a strong distributive law would hold. This cannot hold for parsers without total backtracking and as such seems like it was perhaps erroneously added to cats Alternative law testing. I added a comment linking to some past difficulties cats had with a related issue and also to a discussion that some Haskell Alternative instances are not distributive.

cc @non


non (approved):
Looks good. I think it's great to be explicit about the kind of distributivity we can (or can't) have.

PR #44, opened 2020-11-05
Fix unclear SBT example in documentation

stephenjudkins:


regadas (commented):
@stephenjudkins Thanks for this!

Technically this is not a typo, this how one would add it as a scala.js dep. Maybe we could reword this a bit so we could fit both?

regadas:
```scala
// use this snippet for the JVM
libraryDependencies += "org.typelevel" %% "cats-parse" % "0.0.1"

// use this snippet for JS, or cross-building
libraryDependencies += "org.typelevel" %%% "cats-parse" % "0.0.1"
```

stephenjudkins:
Yeah, I'm not certain what the right way would be here, but I would personally err on just using `%%` since that's what most Scala users (who aren't using scala.js) will default to. Your call!

johnynek:
@stephenjudkins do you want to update the PR with @regadas suggestion? I agree it would be helpful for people less familar with these sbt esoterica.

PR #45, opened 2020-11-05
Add ability to repeat into different structures

johnynek:
I chatted with this idea with @non after seeing an example CSV parser he wrote and seeing where it could have come in handy.

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Accumulator.scala
+++ b/core/shared/src/main/scala/cats/parse/Accumulator.scala
@@ -24,6 +24,8 @@ package cats.parse
 import cats.data.{NonEmptyList, NonEmptyVector}
 import scala.collection.mutable.Builder
 
+/** A limited Builder-like value we use for portability
+  */
 trait Appender[-A, +B] {
   def append(item: A): this.type
   def finish(): B
@@ -63,10 +65,28 @@ object Appender {
 
       def finish(): B = bldr.result()
     }
+
+  val unitAppender: Appender[Any, Unit] =
+    new Appender[Any, Unit] {
+      def append(item: Any) = this
+      def finish(): Unit = ()
+    }
 }
 
-trait Accumulator[-A, +B] {
+/** Creates an appender given the first item to be added
+  * This is used to build the result in Parser1.repAs1
+  */
+trait Accumulator1[-A, +B] {
+  def newAppender(first: A): Appender[A, B]
+}
+
+/** Creates an appender
+  * This is used to build the result in Parser1.repAs
+  */
+trait Accumulator[-A, +B] extends Accumulator1[A, B] {
   def newAppender(): Appender[A, B]
+  def newAppender(first: A): Appender[A, B] =
+    newAppender().append(first)
 }
 
 object Accumulator {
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/cats/parse/Accumulator.scala
```diff
@@ -0,0 +1,145 @@
...
+    new Accumulator[Char, String] {
+      def newAppender() = Appender.charStringAppender()
+    }
+
+  implicit val stringAccumulator: Accumulator[String, String] =
+    new Accumulator[String, String] {
+      def newAppender() = Appender.stringAppender()
+    }
+
+  implicit def listAccumulator[A]: Accumulator[A, List[A]] =
```
non, line 83: Should we have something like:

```scala
def fromBuilder[A, B](makeBuilder: () => Builder[A, B]): Accumulator[A, B] = ???
```

You could use that for both `listAccumulator` and `vectorAccumulator` and anyone else who wants to use `Appender.fromBuilder` will also want something like this.

non: Also what about:

```scala
implicit def monoidAccumulator[A](implicit ev: Monoid[A]): Accumulator[A, A] = ???
```

(And the same for `Semigroup` with `Accumulator1`.)

core/shared/src/main/scala/cats/parse/Accumulator.scala
```diff
@@ -0,0 +1,145 @@
...
+      def append(item: A) = {
+        bldr += item
+        this
+      }
+
+      def finish(): B = bldr.result()
+    }
+}
+
+trait Accumulator[-A, +B] {
```
non, line 68: It seems like having `Accumulator` extend `Accumulator1` would make sense, since:

```scala
def newAppender(first: A): Appender[A, B] = {
  val appender = newAppender()
  appender.append(first)
  appender
}
```

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -592,6 +592,19 @@ object Parser extends ParserInstances {
       softProduct01(parser, void1(that)).map(_._1)
   }
 
+  /** Methods with complex variance type signatures due to covariance.
+    */
+  implicit final class Parser1Methods[A](private val self: Parser1[A]) extends AnyVal {
+    def repAs[B](implicit acc: Accumulator[A, B]): Parser[B] =
+      Parser.repAs(self)(acc)
+
+    def rep1As[B](implicit acc: Accumulator1[A, B]): Parser[B] =
```
non, line 601: Bikeshedding but maybe `repAs1` is better since the 1 is more obvious as a suffix (and that's where we put it in other cases).

core/shared/src/main/scala/cats/parse/Accumulator.scala
non: Lastly, I can imagine arguments for and against including `Set`. The argument for is that often users may want to end up with a set, and if their language ignores duplicates the behavior will be correct. The argument against is that information would be lost, and maybe the user should have to do that themselves. I could imagine building a `MultiSet[A]` (e.g. `Map[A, Int]`) although that seems kinds of loony from a documentation standpoint.

johnynek: These are good questions, but I left them undone for now:

1. I think most use cases will be the ones I made, and I didn't want to add an indirection through one more lambda.
2. As we talked offline, I can't see a good use case and plenty of weird cases for Monoid or Semigroup. For now, I'd rather punt till we have clearer examples. I don't know if it will be much more efficient than mapping and using combine by hand, which also has the benefit of reading more clearly

core/shared/src/main/scala/cats/parse/Accumulator.scala
johnynek: good call. I did this.

core/shared/src/main/scala/cats/parse/Parser.scala
johnynek: yeah, I think your suggestion is right.

non (approved):
Look great! 👍

PR #46, opened 2020-11-06
Case-insensitive string parsers

stephenjudkins:
Some grammars, like SQL, unfortunately need case-insensitive token matching. This adds this functionality.

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -611,13 +611,21 @@ object Parser extends ParserInstances {
   def pure[A](a: A): Parser[A] =
     Impl.Pure(a)
 
+  /** Parse a given string, in a case-insensitive manner,
+    * or fail. This backtracks on failure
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -611,13 +611,21 @@ object Parser extends ParserInstances {
   def pure[A](a: A): Parser[A] =
     Impl.Pure(a)
 
+  /** Parse a given string, in a case-insensitive manner,
+    * or fail. This backtracks on failure
+    * this is an error if the string is empty
+    */
+  def stringCI1(str: String): Parser1[Unit] =
+    if (str.length == 1) charIn(str.charAt(0).toLower, str.charAt(0).toUpper).void
```
johnynek, line 619: nit, can we call charAt once and capture in a val? It makes it a bit easier to read.

johnynek:
This is great, one minor nit. If you have time, follow up with a PR, but in the spirit of merge often, I'll merge this.

PR #50, opened 2020-11-07
Add some ignoreCase methods, rename stringCI to ignoreCase

johnynek:
follow up to #46 

What do you think @stephenjudkins ? I want to publish a new version, but the name just wasn't sitting well with me.

Can you review this and the other new methods I added?

There is one ambiguity: string can be converted to `Iterable[Char]` in scala, so I added `ignoreCaseChar` and `ignoreCaseCharIn` but to save space for the common case, I used ignoreCase to accept a string. A clearer, but more verbose approach would be to do ignoreCaseString, but I feel that's a bit long. That said, ignoring case isn't that common, so maybe that's okay. I don't know.

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -615,11 +615,18 @@ object Parser extends ParserInstances {
     * or fail. This backtracks on failure
     * this is an error if the string is empty
     */
-  def stringCI1(str: String): Parser1[Unit] =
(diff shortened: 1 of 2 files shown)


stephenjudkins:
👍 i think this all looks great. I picked `stringCI` simply because it was the name use in atto but I like these a bit better. I wouldn't quibble too much about `ignoreCaseString` vs `ignoreCase`, I think having string be the "default" thing is probably what 99% of parsers will be doing

PR #51, opened 2020-11-08

regadas:
This is a kind reminder to run `sbt +prePR` and commit the changed files, if any, before submitting.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ import Dependencies._
 addCommandAlias("fmt", "; scalafmtAll; scalafmtSbt")
 addCommandAlias("fmtCheck", "; scalafmtCheckAll; scalafmtSbtCheck")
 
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -4,6 +4,8 @@ import Dependencies._
 addCommandAlias("fmt", "; scalafmtAll; scalafmtSbt")
 addCommandAlias("fmtCheck", "; scalafmtCheckAll; scalafmtSbtCheck")
 
+addCommandAlias("prePR", "fmt")
```
johnynek, line 7: Should we add test into that list? And maybe bench/compile?

johnynek (commented):
Thanks for taking this!

build.sbt
regadas: yup let's do it!

johnynek (approved):
This is awesome, thanks.

PR #52, opened 2020-11-08
Add common Functor#as directly

regadas:


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -198,6 +198,16 @@ sealed abstract class Parser[+A] {
   def flatMap[B](fn: A => Parser[B]): Parser[B] =
     Parser.flatMap(this)(fn)
 
+  /** Replaces parsed values with the given value.
+    */
+  def as[B](b: B): Parser[B] =
+    Parser.as(this)(b)
```
johnynek, line 204: I think here we can just do `.void.map(_ => b)` since this is not fundamental. And remove from the companion.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -321,6 +331,16 @@ sealed abstract class Parser1[+A] extends Parser[A] {
...
 
+  /** This method overrides `Parser#as` to refine the return type.
+    */
+  override def as[B](b: B): Parser1[B] =
+    Parser.as1(this)(b)
+
+  /** This method override `Parser#replicateA` to refine return type.
+    */
+  override def replicateA(n: Int): Parser1[List[A]] =
+    Parser.replicateA1(n, this)
```
johnynek, line 342: This isn't a Parser1 if n == 0, right?

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -321,6 +331,16 @@ sealed abstract class Parser1[+A] extends Parser[A] {
   override def flatMap[B](fn: A => Parser[B]): Parser1[B] =
     Parser.flatMap10(this)(fn)
 
+  /** This method overrides `Parser#as` to refine the return type.
+    */
+  override def as[B](b: B): Parser1[B] =
+    Parser.as1(this)(b)
```
johnynek, line 337: I think void.map is fine.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -854,6 +874,30 @@ object Parser extends ParserInstances {
   def flatMap01[A, B](pa: Parser[A])(fn: A => Parser1[B]): Parser1[B] =
     Impl.FlatMap1(pa, fn)
 
+  /** Replaces parsed values with the given value.
+    */
+  def as[A, B](pa: Parser[A])(b: B): Parser[B] =
+    map(Impl.Void(pa))(_ => b)
```
johnynek, line 880: Actually def void does more than this. It also unmaps pa so we can simplify it. We should use that but I think as I mentioned just using the void method is the way to go.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1645,6 +1691,95 @@ object Parser extends ParserInstances {
...
+          if (replicateCapture(p1, n, state, app)) app.finish()
+          else ignore
+        } else {
+          replicateNoCapture(p1, n, state)
+          ignore
+        }
+      }
+    }
+
+    case class Replicate1[A, B](p1: Parser1[A], n: Int, acc: Accumulator1[A, B])
```
johnynek, line 1766: What do you think of this idea instead: we take a min and a max in Rep, and for replicate then min is the max. But this also enables a new combinator: parse at most n things. Of course that is a bit of a weird combinator so maybe not with special casing.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -854,6 +874,30 @@ object Parser extends ParserInstances {
...
+    */
+  def replicateA[A](n: Int, pa: Parser[A])(implicit
+      acc: Accumulator[A, List[A]]
+  ): Parser[List[A]] =
+    Impl.Replicate(pa, n, acc)
+
+  /** Apply Parser1 `n` times
+    */
+  def replicateA1[A](n: Int, pa: Parser1[A])(implicit
+      acc: Accumulator1[A, List[A]]
```
johnynek, line 897: I think we should just pass the List accumulator and not use an implicit here since it compilcates the call site but does not add power. What do you think?

Alternatively: we make this replicateAs and then allow us to replicate into different types (like Vector)

johnynek (commented):
I had a few comments to think about.

One thing: I think the default implementation of replicateA (which uses sequence) might actually be fine in terms of performance and keeps the internals simpler.

johnynek:
Thanks for taking this on!

johnynek:
The more I think about it, replicateA is a really bad name. The A comes from the fact that replicate is already a method on collections, IIRC. So, the A stands for applicative. I think maybe .exactly or something might be a better name. I used rep, following fastparse, to mean repeat at least, which is the most common kind of repetition (and almost always you want to repeat 0 or more times, or 1 or more times)...

honestly, maybe we should just punt on the whole replicate question and remove that for now. I can't even imagine a use case at the moment, and if you want it, you can use replicateA from Applicative to get it.

What do you think?

core/shared/src/main/scala/cats/parse/Parser.scala
regadas: Yeah you are right, it's better to use it.

regadas:
@johnynek regarding `replicateA` from name to use case it's a bit weird. Most of the time what we want is really 0/1 or more times so I'm fine in let it rest for now.

I'll clean it up in a bit!

johnynek:
Thanks a ton!

I'm going to make one more PR than release a new version today.

johnynek (approved):

PR #58, opened 2020-11-11
Add a few new laws and optimizations

johnynek:
A few more laws occurred to me so I added them.

At the same time, I realized we could leverage some of those properties when statically building up the parsers to make them a bit simpler.

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -825,6 +825,8 @@ object Parser extends ParserInstances {
         Impl.Map(p0, AndThen(f0).andThen(fn))
       case Impl.Map1(p0, f0) =>
         Impl.Map1(p0, AndThen(f0).andThen(fn))
+      case Impl.Fail() => Impl.Fail()
+      case Impl.FailWith(str) => Impl.FailWith(str)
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -834,6 +837,8 @@ object Parser extends ParserInstances {
     p match {
       case Impl.Map1(p0, f0) =>
         Impl.Map1(p0, AndThen(f0).andThen(fn))
+      case Impl.Fail() => Impl.Fail()
+      case Impl.FailWith(str) => Impl.FailWith(str)
```
rossabaker, line 841: Can you reuse the same instances with a cast here?

johnynek: I could, but my heuristic is that paser build time is cheap (done once), but parser run time is where we want to optimize, so I didn't bother... but why not...

johnynek:
took your suggestion @rossabaker

core/shared/src/main/scala/cats/parse/Parser.scala
rossabaker: :laughing:, I guess we each persuaded the other. I'm happy either way.

Issue #64, opened 2020-11-14
Consistency of return types


rossabaker:
Spun off from an [http4s issue](https://github.com/http4s/http4s/pull/3855#discussion_r522666228).

* `char` returns a `Parser1[Unit]`. We know what it captured, so we don't return it.
* `charIn` returns a `Parser1[Char]`. It captured one of a set of characters, and we want to know which.
* `ignoreChar` returns a `Parser1[Unit]`. Arguably like `charIn` in that we don't know what it cpatured, but the docs tell us to call `.string` if we need the result.
* `string1` returns a `Parser[Unit]`.  Seems consistent with `.char`.

Did we get this right, or should everything return what it captured?

johnynek:
A few notes:

1. Returning the captured String from string is free. We don't have to allocate (we know what we expect already).
2. Returning the string in ignoreCase is not free, we need to allocate a substring if we are capturing. If we aren't capturing there is an additional Boolean check we don't have now, probably not a big deal.
3. Returning the Char in char is cheap since we have the char, there is likely a boxing of the char but I assume the JVM has a cache of those so we are hitting a lookup table that might even be in the CPU cache.

To summarize, I doubt this is a perfect question to add the return values and since we have using `*>` and void as recommended style I guess it doesn't do much harm either.

On the other hand, matching a constant string and wanting that exact thing back is rare IME. Usually you combine it a bit more then capture a larger block with string or you use .as to change it. So I haven't felt the need.

I think we are consistent now (except arguably the ignoreCase variants maybe should return what they capture), so I think the discussion would be helped if we had some examples where one did want to capture directly the results of these parsers.

rossabaker:
As I get deeper into this, I'm feeling more comfortable with the way it is.  A mild discomfort: in the spirit of "return Unit if constant, return the value when variable", this rule from RFC5234 should return a `Char`. To do so would require a couple annoying `.as`es, because it's composed of two Unit rules.

```
  val wsp: Parser1[Unit] =
    sp.orElse1(htab)
```

I made a judgment call that which kind of whitespace is found is generally uninteresting, and left it as `Unit`.

PR #66, opened 2020-11-14
RFC5234 common rules

### LINKED ISSUE #65: Provide RFC5234 core rules?
There are [many RFCs](https://rfc-wiki.org/index.php?title=Special:WhatLinksHere/RFC5234&limit=500&from=0) that reference the [core rules of RFC5234](https://tools.ietf.org/html/rfc5234#appendix-B.1). Is there any interest in an object that provides those?  http4s has already implemented those imported by RFC7230.

rossabaker:
Fixes #65.

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Rfc5234.scala
+++ b/core/shared/src/main/scala/cats/parse/Rfc5234.scala
@@ -64,7 +64,7 @@ object Rfc5234 {
   /** `0` to `9`
     */
   val digit: Parser1[Char] =
-    Parser.charIn('0' to '9')
+    Numbers.digit
 
   /** double quote (`"`)
     */
--- a/core/shared/src/main/scala/cats/parse/Rfc5234.scala
+++ b/core/shared/src/main/scala/cats/parse/Rfc5234.scala
@@ -59,7 +59,7 @@ object Rfc5234 {
 
   /** controls */
   val ctl: Parser1[Char] =
-    Parser.charIn(0x7f, (0x00.toChar to 0x1f.toChar: _*))
+    Parser.charIn(0x7f, (0x00.toChar to 0x1f.toChar): _*)
 
   /** `0` to `9`
     */
--- a/core/shared/src/main/scala/cats/parse/Rfc5234.scala
+++ b/core/shared/src/main/scala/cats/parse/Rfc5234.scala
@@ -59,7 +59,7 @@ object Rfc5234 {
 
   /** controls */
   val ctl: Parser1[Char] =
-    Parser.charIn(0x00.toChar to 0x1f.toChar).orElse1(Parser.char(0x7f).as(0x7f))
+    Parser.charIn(0x7f, (0x00.toChar to 0x1f.toChar: _*))
 
   /** `0` to `9`
     */
@@ -99,7 +99,7 @@ object Rfc5234 {
     * other contexts.
     */
   val lwsp: Parser[Unit] =
-    Parser.rep(wsp.orElse1((crlf ~ wsp).void)).void
+    Parser.rep(wsp.orElse1(crlf *> wsp)).void
 
   /** 8 bits of data
     */
--- a/core/shared/src/main/scala/cats/parse/Rfc5234.scala
+++ b/core/shared/src/main/scala/cats/parse/Rfc5234.scala
@@ -103,11 +103,11 @@ object Rfc5234 {
 
   /** 8 bits of data
     */
-  val octet: Parser[Char] =
+  val octet: Parser1[Char] =
     Parser.charIn(0x00.toChar to 0xff.toChar)
 
   /** visible (printing) characters
     */
-  val vchar: Parser[Char] =
+  val vchar: Parser1[Char] =
     Parser.charIn(0x21.toChar to 0x7e.toChar)
 }
--- a/core/shared/src/main/scala/cats/parse/Rfc5234.scala
+++ b/core/shared/src/main/scala/cats/parse/Rfc5234.scala
@@ -1,0 +1,113 @@
+/*
+ * Copyright (c) 2020 Typelevel
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package cats.parse
+
+/** Parsers for the common rules of RFC5234. These rules are
+  * referenced by several RFCs.
+  *
+  * @see [[https://tools.ietf.org/html/rfc5234]]
+  */
+object Rfc5234 {
+
+  /** A-Z and a-z, without diacritics
+    */
+  val alpha: Parser1[Char] =
+    Parser.charIn('A' to 'Z').orElse1(Parser.charIn('a' to 'z'))
+
+  /** `0` or `1`
+    */
+  val bit: Parser1[Char] =
+    Parser.charIn('0' to '1')
+
+  /** any 7-bit US-ASCII character, excluding NUL
+    */
+  val char: Parser1[Char] =
+    Parser.charIn(0x01.toChar to 0x7f.toChar)
+
+  /** carriage return
+    */
+  val cr: Parser1[Unit] =
+    Parser.char('\r')
+
+  /** linefeed
+    */
+  val lf: Parser1[Unit] =
+    Parser.char('\n')
+
+  /** Internet standard newline */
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/parse/Rfc5234.scala
```diff
@@ -0,0 +1,113 @@
...
+object Rfc5234 {
+
+  /** A-Z and a-z, without diacritics
+    */
+  val alpha: Parser1[Char] =
+    Parser.charIn('A' to 'Z').orElse1(Parser.charIn('a' to 'z'))
+
+  /** `0` or `1`
+    */
+  val bit: Parser1[Char] =
```
rossabaker, line 38: Was tempted to make this return a `Boolean`.  Similar considerations for `digit` and `hexdigit`.  Maybe it's good to have these raw versions, and maps for the obvious interpretations.

core/shared/src/main/scala/cats/parse/Rfc5234.scala
```diff
@@ -0,0 +1,113 @@
...
+  val lf: Parser1[Unit] =
+    Parser.char('\n')
+
+  /** Internet standard newline */
+  val crlf: Parser1[Unit] =
+    Parser.string1("\r\n")
+
+  /** controls */
+  val ctl: Parser1[Char] =
+    Parser.charIn(0x00.toChar to 0x1f.toChar).orElse1(Parser.char(0x7f).as(0x7f))
```
rossabaker, line 62: That `as` is awkwardness to consider in #64.

core/shared/src/main/scala/cats/parse/Rfc5234.scala
```diff
@@ -0,0 +1,113 @@
...
+    *
+    * Use of this rule permits lines containing only white space that
+    * are no longer legal in mail headers and have caused
+    * interoperability problems in other contexts.
+    *
+    * Do not use when defining mail headers and use with caution in
+    * other contexts.
+    */
+  val lwsp: Parser[Unit] =
+    Parser.rep(wsp.orElse1((crlf ~ wsp).void)).void
```
rossabaker, line 102: Threw enough `.voids` at it to stop the warnings.

core/shared/src/main/scala/cats/parse/Rfc5234.scala
```diff
@@ -0,0 +1,113 @@
...
+    */
+  val htab: Parser1[Unit] =
+    Parser.char('\t')
+
+  /** space */
+  val sp: Parser1[Unit] =
+    Parser.char(' ')
+
+  /** white space (space or horizontal tab) */
+  val wsp: Parser1[Unit] =
```
rossabaker, line 89: In the spirit of "return a value if it can be more than one", maybe this should be a `Parser1[Char]`. Another #64 deliberation.

rossabaker (commented):
This probably deserves a few tests, but wanted to post it as it informs #64.

core/shared/src/main/scala/cats/parse/Rfc5234.scala
johnynek: Or `charIn(0x7f, (0x0.toChar to 0x1f.toChar) : _*)` might be simpler, or build a List.

core/shared/src/main/scala/cats/parse/Rfc5234.scala
johnynek: `crlf *> wsp` will rid you of one of those void calls.

That operator is your friend with parsers.

johnynek (commented):
Made some comments. Use your judgement if you want to take them. I don't think any make a performance difference.

One thing I would like is to add some basic tests that these things match what they say. A bit tedious, because they are so simple it is more work to make the test.

One pattern would be to make a Gen for each Parser, then the law is that everything parses. That said, we are just encoding things twice, once in the Gen and once in the Parser. It's not as great as a real law, but at least we have two chances to catch any errors.

core/shared/src/main/scala/cats/parse/Rfc5234.scala
rossabaker: Yes, nice. Still reacclimating myself from parboiled's magic stack.

core/shared/src/main/scala/cats/parse/Rfc5234.scala
```diff
@@ -0,0 +1,113 @@
...
+    * Do not use when defining mail headers and use with caution in
+    * other contexts.
+    */
+  val lwsp: Parser[Unit] =
+    Parser.rep(wsp.orElse1(crlf *> wsp)).void
+
+  /** 8 bits of data
+    */
+  val octet: Parser[Char] =
+    Parser.charIn(0x00.toChar to 0xff.toChar)
```
johnynek, line 107: Will this really work? Like if you have random bytes in the middle, but it was decoded might that not decoding break some of these octet sequences?

Can we add a comment about this issue?

core/shared/src/test/scala/cats/parse/Rfc5234Test.scala
```diff
@@ -0,0 +1,110 @@
...
+  singleMultiCharProperties("bit", Rfc5234.bit, Set('0', '1'))
+  singleMultiCharProperties("char", Rfc5234.char, Set(0x01.toChar to 0x7f.toChar: _*))
+  singleConstCharProperties("cr", Rfc5234.cr, 0x0d.toChar)
+  singleMultiCharProperties("ctl", Rfc5234.char, Set(0x01.toChar to 0x1f.toChar: _*) + 0x7f.toChar)
+  singleMultiCharProperties("digit", Rfc5234.digit, Set(0x30.toChar to 0x39.toChar: _*))
+  singleConstCharProperties("dquote", Rfc5234.dquote, 0x22.toChar)
+  singleMultiCharProperties(
+    "hexdig",
+    Rfc5234.hexdig,
+    Set('0' to '9': _*) ++ Set('A' to 'F': _*) ++ Set('a' to 'F': _*)
```
johnynek, line 70: I think you didn't mean a capital F in the second one right? So I'm surprised this passed. It should fail right?

core/shared/src/test/scala/cats/parse/Rfc5234Test.scala
```diff
@@ -0,0 +1,110 @@
...
+  singleConstCharProperties("sp", Rfc5234.sp, 0x20.toChar)
+  singleMultiCharProperties("vchar", Rfc5234.vchar, Set(0x21.toChar to 0x7e.toChar: _*))
+  singleCharProperties("wsp", Rfc5234.wsp, Set(0x20.toChar, 0x09.toChar), _ => ())
+
+  test("crlf accepts \r\n") {
+    assertEquals(Rfc5234.crlf.parseAll("\r\n"), Right(()))
+  }
+  property("crlf rejects all but \r\n") {
+    forAll { s: String =>
+      assert(Rfc5234.crlf.parseAll(s).isLeft)
```
johnynek, line 84: Aren't you just getting lucky here? Why isn't s the matching string?

core/shared/src/test/scala/cats/parse/Rfc5234Test.scala
```diff
@@ -0,0 +1,110 @@
...
+  property("lwsp accepts all linear white space") {
+    val genWsp = Gen.oneOf(0x20.toChar, 0x09.toChar)
+    val genLwsp = Gen
+      .listOf(
+        Gen.oneOf(
+          genWsp.map(_.toString),
+          genWsp.map("\r\n" + _)
+        )
+      )
+      .map(_.mkString(""))
```
johnynek, line 97: Isn't mkString without an arg the same as passing empty string?

johnynek (commented):
Didn't look in detail but seems the dotty test failed.

johnynek:
Really nice tests, BTW. Great work indeed!

core/shared/src/main/scala/cats/parse/Rfc5234.scala
```diff
@@ -0,0 +1,113 @@
...
+    Parser.string1("\r\n")
+
+  /** controls */
+  val ctl: Parser1[Char] =
+    Parser.charIn(0x7f, (0x00.toChar to 0x1f.toChar: _*))
+
+  /** `0` to `9`
+    */
+  val digit: Parser1[Char] =
+    Parser.charIn('0' to '9')
```
regadas, line 67: Outside of this PR; but I was just wondering if we could explore the idea of reusable `Parser` i.e this one is already available under `Numbers#digit`.

I see some other defined here that could be useful for other use cases.

core/shared/src/main/scala/cats/parse/Rfc5234.scala
rossabaker: This is a good question. We're caught between RFCs written for protocols that assume chars are single bytes, and a UTF-8 world.  The rule is generally appears in binary contexts.  In the HTTP specs, the rule is only used for the message body, which isn't a good domain for cats-parse. And its use of 0x80-0xff is constrained to "obs[olete] text", which is basically, "our forebears only knew ASCII. Good luck."

I like to be complete, but I think providing this one encourages dangerous use. I'd be content to just remove it.

core/shared/src/test/scala/cats/parse/Rfc5234Test.scala
rossabaker: Great catch. This is a time bomb. `Set('a' to 'F': _*)` is empty.  The positive test passes on its incomplete alphabet, and the negative test only had a 6 in (Char.MaxValue - 16) chance of failing per generation.  The uniform distribution of `genInvalid` could be improved.

core/shared/src/test/scala/cats/parse/Rfc5234Test.scala
rossabaker: I don't understand "Why isn't s the matching string", but I meant to filter out the (unlikely) `"\r\n"`.

core/shared/src/test/scala/cats/parse/Rfc5234Test.scala
rossabaker: `mkString()` steered me to the 1-arg and 3-arg versions. I shouldn't have put the parens on a pure call.

core/shared/src/main/scala/cats/parse/Rfc5234.scala
rossabaker: I was trying to hew closely to the spec, but the test now does so.  We could use the `Numbers` version.

core/shared/src/test/scala/cats/parse/Rfc5234Test.scala
rossabaker: Improving that generator exposed this, and caught two additional bugs in my test!

johnynek (approved):
This looks great. Thanks for the extensive tests!

rossabaker:
Do we want to remove that `octet` rule before the next release?  If not, I'll follow up with some prose.

PR #68, opened 2020-11-14
mergeCharIn tailrec

regadas:


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1775,32 +1775,31 @@ object Parser extends ParserInstances {
...
+            result ::: AnyChar.asInstanceOf[P] :: tail1
           case CharIn(m, bs, _) :: tail =>
-            loop(tail, (m, bs) :: front)
+            loop(tail, (m, bs) :: front, result)
           case h :: tail =>
             // h is not an AnyChar or CharIn
             // we make our prefix frontRes
             // and resume working on the tail
-            frontRes ::: (h :: loop(tail, Nil))
+            loop(tail, Nil, result ::: frontRes ::: h :: Nil)
```
johnynek, line 1798: Concat on the right is O(N) on a list so this is quadratic no? If we say, the list isn't long, that also means it won't blow the stack. So in the case this change helps us it also causes N^2 complexity I think.

Can we use a Chain to get fast concat and then convert to a List at the end? Or append on the left and reverse at the end?

What do you think? Does that sound right?

regadas: Right, got to the same conclusion as well. Updated it with `Chain` as it seems to fit this better.

johnynek (approved):
This is great, thanks!

PR #74, opened 2020-11-18
Fix issue 72

### LINKED ISSUE #72: Failing test cases in main
I'm seeing two tests failing:

 - `cats.parse.ParserTest.voided only changes the result`
 - `cats.parse.ParserTest.with1 *> and with1 <* work as expected`

The errors have the form:

```
values are not the same
=> Diff (- obtained, + expected)
           upper = 'ﷷ'
+        ),
+        Fail(
+          offset = 0
         )
```

To reproduce add the following to `ParserTest`:

```scala
override val scalaCheckInitialSeed = "SDzb3fKPxR67aeO2sgq4BlvTm5NphF9OM4j-dSIS9RD="
```

I haven't dug into this -- just figured I should report it ASAP.

johnynek:
close #72 

This is a matter of optimizations creating redundant failure offsets. Originally I removed those, but then I took that out to make things simpler.

Now I go back to removing a naked Fail Expectation which is at an offset that other non Fail messages have already failed at. This adds nothing to the user.

We have to have a naked Fail expectation to have the empty value for the MonoidK, but users will probably prefer failWith which allows them to set a custom error message, which is not normalized away.

cc @non

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -467,9 +467,11 @@ object Parser extends ParserInstances {
       * This is called just before finally returning an error in Parser.parse
       */
     def unify(errors: NonEmptyList[Expectation]): NonEmptyList[Expectation] = {
+      val el = errors.toList
+
       // merge all the ranges:
       val rangeMerge: List[InRange] =
-        errors.toList
+        el
           .collect { case InRange(o, l, u) => (o, l to u) }
           .groupBy(_._1)
           .iterator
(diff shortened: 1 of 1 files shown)


johnynek:
Note, the performance of Error normalization isn't great since we do a few passes over the list. We could make it faster, but note this is only called once at the outer most parse and only in the event that parsing failed, which presumably isn't the common case.

So, I haven't been too uptight about making sure this is fast.

A faster implementation would be to make an array of a fixed size of all errors originally, then mutate that array in place as we are normalizing and sorting, then finally make a single NonEmptyList at the end.

But I think we have no benchmarks that even measure the failing case, so it wouldn't yet speed anything up.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -483,9 +485,20 @@ object Parser extends ParserInstances {
           }
           .toList
 
-      if (rangeMerge.isEmpty) errors.distinct.sorted
+      // we don't need Fails that point to duplicate offsets
+      val nonFailOffsets: Set[Int] =
+        el.iterator.filterNot(_.isInstanceOf[Fail]).map(_.offset).toSet
+
+      val errors1 = NonEmptyList.fromListUnsafe(
```
johnynek, line 492: note, this still has to have at least one item, if there were only one item, there can't be a duplicate, so nothing is filtered out.

regadas (approved):
👍 

@johnynek even though hitting this will probably not be that common, I think we could still address the perf subject afterwards because I got the feeling it will also make it somewhat cleaner

PR #75, opened 2020-11-18
bump github actions, regenerate ci.yml

johnynek:
replaces #71 

we needed to rerun to build ci.yml.


mpilquist:
Note scala steward is now able to auto regenerate the github workflow as part of the upgrade PR. So we shouldn't need to do this anymore.

johnynek:
Cool!

johnynek:
I need to learn how to add custom rules like that. Are they ad-how or is there some system for it?

fthomas:
@johnynek It will do it automatically on the next update of sbt-github-actions. There is nothing you need to do to enable it. The PR will then look like this https://github.com/scala-steward-org/test-repo-2/pull/10.

regadas:
These hooks are added directly to `scala-steward`

https://github.com/scala-steward-org/scala-steward/blob/928dbc6c4aa0c8a6aef19af027ec7112700b8163/modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala#L72-L80

johnynek:
@fthomas That's great! but I mean in general, there are libraries or tools that need some special migration. I'm more interested in how those kinds of things are added in the general case.

johnynek:
looks like it is still ad-hoc, but you can imagine declaring in some jar some code that should be run to update the repo... but it is challenging to imagine the security issues, etc...

fthomas:
Right, it would be great if libraries and plugins would ship with this information so that multiple tools could use it and more maintainers would provide it. The same applies for Scalafix migrations. My hope is that the experience in Scala Steward with post-update hooks and Scalafix migrations can help designing such facility.

djspiewak:
An example is that I would really like to *also* get this functionality for sbt-spiewak upgrades (since it brings in sbt-github-actions as an upstream dependency), but it would be unreasonable for scala-steward to hard-code all of that for increasingly niche plugins. Better if there were some way for me to publish it ourselves.

fthomas:
> it would be unreasonable for scala-steward to hard-code all of that for increasingly niche plugins

I don't mind adding hooks for niche plugins to Scala Steward until there is a general solution. If you want to run `sbt githubWorkflowGenerate` on sbt-spiewak updates, just add a `PostUpdateHook` like [this](https://github.com/scala-steward-org/scala-steward/blob/928dbc6c4aa0c8a6aef19af027ec7112700b8163/modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala#L72-L80).

> Better if there were some way for me to publish it ourselves.

Agreed. I'll open an issue in scala-steward for further discussion.

PR #76, opened 2020-11-19
Avoid oneOf reverse on List

regadas:
Just noticed this.

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -25,6 +25,7 @@ import cats.{Eval, Monad, Defer, Alternative, FlatMap, Now, MonoidK, Order}
 import cats.data.{AndThen, Chain, NonEmptyList}
 
 import cats.implicits._
+import scala.collection.mutable.ListBuffer
 
 /** Parser[A] attempts to extract an `A` value from the given input,
   * potentially moving its offset forward in the process.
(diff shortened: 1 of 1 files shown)


johnynek:
I don't love passing mutable variables through a recursive loop. I'd rather either go all in and make a while loop to signal to the reader that we are in an imperative mode here, or just capture the buffer to make it clear there is only ever one mutable ListBuffer.

I'll merge this since you took the time, but I think it is worth thinking of three tiers of performance concerns:

1. cost to build parsers.
2. cost to run parsers.
3. cost to handle errors.

My belief is that it is 2 that by far dominates and that we shouldn't sacrifice much clarity and immutability in 1 or 3. On point 2 we should save our full budget for crazy tricks to maximize the performance of running (typically correct) parsers. On 3, note, when a parser fails, generally something will be reported to a user, and that will be very slow. So, a bit of slowness there if it results in clearer code, is probably fine.

johnynek (approved):
Added some comments, but I will merge this.

regadas:
You convinced me with:
> I don't love passing mutable variables through a recursive loop. I'd rather either go all in and make a while loop to signal to the reader that we are in an imperative mode here, or just capture the buffer to make it clear there is only ever one mutable ListBuffer.

It's a very good argument and I totally agree with you here.

> I'll merge this since you took the time

I would be totally fine in closing this 😄  I value more reaching a good consensus and understanding on how do we want things to be.

Thank you!

johnynek:
It's fine. I appreciate your PRs and all the help you've given.

PR #77, opened 2020-11-21
Update to cats-core-2.3.0-M2

rossabaker:
The Dotty build is pulling in cats-2.2.0_2.13, and we can't mix _2.13 and _3.0.0-M1 dependencies.  cats-2.3.0-M2 is fully cross-published for Dotty.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -2,7 +2,7 @@ import sbt._
 import org.portablescala.sbtplatformdeps.PlatformDepsPlugin.autoImport._
 
 object Dependencies {
-  lazy val cats = Def.setting("org.typelevel" %%% "cats-core" % "2.2.0")
+  lazy val cats = Def.setting("org.typelevel" %%% "cats-core" % "2.3.0-M2")
   lazy val munit = Def.setting("org.scalameta" %%% "munit" % "0.7.18")
   lazy val munitScalacheck = Def.setting("org.scalameta" %%% "munit-scalacheck" % "0.7.18")
   lazy val scalaCheck = Def.setting("org.scalacheck" %%% "scalacheck" % "1.14.3")


rossabaker:
I'm not usually a fan of milestones depending on milestones. Pick one:

1. I get over it. Before 1.0, I can.
2. cats-parse moves to milestones.
3. We make the cats version conditional on the Scala version.

regadas:
@rossabaker tbh not a very big fan of (3) and between (1) and (2) I'm more i'm leaning towards (1).

johnynek:
How far off do you think the non milestone build is?

Since I guess almost no one is using dotty, and it is just to get things building and test out our CIs etc...

I would personally actually prefer option 3 (make the dotty build the only one that depends on the milestone).

That way we can publish a new real version of this, IMO, since dotty is already a pre-release type thing.

regadas:
Well I was checking dotty milestones and the RC1 due date is Dec 18 so 3.0.0 prob in Jan? I'm assuming cats will follow right after.

I've been burned before with the multiple versions before 😄 but you got a point with:

> Since I guess almost no one is using dotty, and it is just to get things building and test out our CIs etc...

so it's probably ok.

Also, how soon do you want to have a "real" version? which issues do you want to see in before doing it?

rossabaker:
I started a conversation on the cats-dev Gitter about the Cats 2.3 release timeline to see if that lights a clearer path forward.

I don't usually like option 3, but as long as Cats retains a good source compatibility story (we know binary will be), it's probably tenable for a couple months.

johnynek:
Keep in mind we are only using a few traits from cats. I'd be shocked if option 3 didn't work out for the short term.

djspiewak:
Fwiw, Cats 2.3.0 final should be released today

johnynek:
Thanks for the note @djspiewak - I think that makes it moot. Just wait for that.

rossabaker:
I'll close this one, but would love a release on cats-2.3.0 before Saturday.  I'm hoping to enlist people to port http4s parsers at the LSUG open source event.

PR #78, opened 2020-11-22
Add SemVer Parser

oguzhanunlu:
A potential part of #9 

I wanted to get myself familiar with parser combinators in general and played with this lib over the weekend.

I tried to implement a SemVer parser as an exercise by following [the BNF grammar of SemVer](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions) and wanted to share for both feedback and contribution if you'd like to incorporate it.

Thanks in advance.

### DIFF
--- a/core/shared/src/main/scala/cats/parse/SemVer.scala
+++ b/core/shared/src/main/scala/cats/parse/SemVer.scala
@@ -1,0 +1,69 @@
+/*
+ * Copyright (c) 2020 Typelevel
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package cats.parse
+
+/** SemVer 2.0.0 Parser based on https://semver.org */
+object SemVer {
+
+  val dot: Parser1[Char] = Parser.charIn('.')
+  val hyphen: Parser1[Char] = Parser.charIn('-')
+  val plus: Parser1[Char] = Parser.charIn('+')
+
+  val letter: Parser1[Char] = Parser.ignoreCaseCharIn('a' to 'z')
+
+  val positiveDigit: Parser1[Char] = Numbers.nonZeroDigit
+
+  val nonDigit: Parser1[Char] = letter orElse1 hyphen
+
+  val identifierChar: Parser1[Char] = Numbers.digit orElse1 nonDigit
+
+  val identifierChars: Parser1[String] = identifierChar.rep1.string
+
+  val numericIdentifier: Parser1[String] = Numbers.nonNegativeIntString
+
+  val alphanumericIdentifier: Parser1[String] = identifierChars
+
+  val buildIdentifier: Parser1[String] = alphanumericIdentifier
+
+  val preReleaseIdentifier: Parser1[String] = alphanumericIdentifier
+
+  val dotSeparatedBuildIdentifiers: Parser1[String] = Parser.rep1Sep(buildIdentifier, 1, dot).string
+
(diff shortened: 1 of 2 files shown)


rossabaker:
Thanks for your interest! Only have time for a drive-by review now, but: would it be more interesting to parse to a semver class from which the major, minor, etc. can be queried, instead of producing a (validated) String?

core/shared/src/main/scala/cats/parse/SemVer.scala
```diff
@@ -0,0 +1,69 @@
...
+  val dotSeparatedPreReleaseIdentifiers: Parser1[String] =
+    Parser.rep1Sep(preReleaseIdentifier, 1, dot).string
+
+  val preRelease: Parser1[String] = dotSeparatedPreReleaseIdentifiers
+
+  val patch: Parser1[String] = numericIdentifier
+  val minor: Parser1[String] = numericIdentifier
+  val major: Parser1[String] = numericIdentifier
+
+  val core: Parser1[String] = (major ~ dot ~ minor ~ dot ~ patch).string
```
johnynek, line 62: If this file imports `cats.implicits._` we can use Apply syntax here I believe:

```
case class Core(major: String, minor: String, patch: String)

val core: Parser1[Core] = (major, dot *> minor, dot *> patch).mapN(Core(_, _, _))
val coreString: Parser1[Core] = core.string
```

core/shared/src/main/scala/cats/parse/SemVer.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  val semver: Parser1[String] =
+    ((core ~ hyphen ~ preRelease ~ plus ~ build).backtrack orElse1
+      (core ~ plus ~ build).backtrack orElse1
+      (core ~ hyphen ~ preRelease).backtrack orElse1
+      core).string
```
johnynek, line 68: this looks correct but uses a bit of backtracking, which isn't ideal for performance or identifying exactly where an error happened.

It is better to factor out the common parts:
1. each orElse branch starts with core: pull that out to the front.
2. then you either have: hyphen, plus, or nothing.... write separate parsers for those cases and you don't need to backtrack:

```scala
case class SemVer(core: Core, prerelease: Option[String], build: option[String])

val semver: Parser1[SemVer] = {
  val withHypen: Parser1[(Option[String], Option[String])] =
    (hypen *> prerelease.map(Some(_)) ~ (plus *> build).?)

  val withPlus: Parser1[(Option[String], Option[String])] =
    plus *> build.map { b => (None, Some(b)) }

  val done: Parser[(Option[String], Option[String])] = Parser.pure((None, None))

  (core ~ Parser.oneOf(withHyphen :: withPlus :: done :: Nil))
    .map { case (s, (p, b)) =>
      SemVer(c, p, b)
    }
}

val semverString: Parser1[String] = semver.string
```

Now, if you just care about the string, it should be just as fast (since calling .string recursively removes all the map operations and removes any internal allocations of lists and just does one final outer string copy on the matched ranges.

Also, the above does not backtrack since it only needs to look at a single character after parsing core to see if it is in the hyphen, plus, or finished case.

Does something like that look right?

core/shared/src/test/scala/cats/parse/SemVerTest.scala
```diff
@@ -0,0 +1,42 @@
...
+import org.scalacheck.Prop.forAll
+import org.scalacheck.Gen
+
+class SemVerTest extends munit.ScalaCheckSuite {
+
+  val genSemVer: Gen[String] = for {
+    major <- Gen.choose(0, 10)
+    minor <- Gen.choose(0, 10)
+    patch <- Gen.choose(0, 100)
+    preRelease <- Gen.oneOf("-alpha", "-beta", "-alpha.1")
```
johnynek, line 33: can we include the empty string in here too? That should work, right?

core/shared/src/test/scala/cats/parse/SemVerTest.scala
```diff
@@ -0,0 +1,42 @@
...
+    buildMetadata <- Gen.oneOf("+001", "+20130313144700", "+exp.sha.5114f85", "+21AF26D3")
```
johnynek, line 34: can we include the empty string in here too?

johnynek (changes requested):
Thank you so much! This is really great.

It is a great example parser, and I'm happy to have it as a built in. Also, as you note, it is a great case for people to look at to learn how to write parsers.

I did find a couple of areas where I suggest a change. Let me know what you think. Although I do think your version is correct, I'd prefer to remove as much backtracking as possible for parsers we merge as examples.

core/shared/src/test/scala/cats/parse/SemVerTest.scala
oguzhanunlu: Right, that should work, I'll add that as well

core/shared/src/test/scala/cats/parse/SemVerTest.scala
oguzhanunlu: Yes, I'll add it too

core/shared/src/main/scala/cats/parse/SemVer.scala
oguzhanunlu: This looks good to me, I'll update my work and get back

oguzhanunlu:
Thanks for the quick & kind feedback @rossabaker @johnynek ! Let me address your feedback and get back shortly

johnynek (approved):
Looks great! Thanks!

oguzhanunlu:
@johnynek , I appreciate the detailed feedback on how to avoid  `backtrack` , I think I could update `core` as following but I went with your solution so that future parser authors can see different ways in this example.

```scala
val core: Parser1[Core] = (major ~ (dot *> minor) ~ (dot *> patch)).map {
  case ((ma, mi), pa) => Core(ma, mi, pa)
}
```

Please let me know if the last commit looks good so that I can squash and force-push to keep the commit history clean.

johnynek:
@oguzhanunlu This looks great. We always squash merge, so there is no need to rebase/squash.

Thanks for your contribution!

oguzhanunlu:
I appreciate you & @rossabaker  taking time to look at my PR. I'll be following the project and try to contribute more if I can.

PR #79, opened 2020-11-22
Add `between` and `surroundedBy` combinators

regadas:


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -388,6 +388,21 @@ sealed abstract class Parser1[+A] extends Parser[A] {
   def rep1(min: Int): Parser1[NonEmptyList[A]] =
     Parser.rep1(this, min = min)
 
+  /** Use this parser to parse between values.
+    *
+    * Parses `b` followed by `this` and `c`.
+    * Returns only the values extracted by `this` parser.
+    */
+  def between[B, C](b: Parser[B], c: Parser[C]): Parser[A] =
```
johnynek, line 396: Can we used `Parser[Any]` here and leverage covariance? That way the signature is simpler (no type params).

johnynek (changes requested):
This is great.

Can we override this for Parser1 since I think these always would return a Parser1 in those cases?

johnynek:
You might also want to try using this in the JSON benchmark.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -388,6 +388,21 @@ sealed abstract class Parser1[+A] extends Parser[A] {
...
+    b *> this <* c
```
johnynek, line 397: One minor item: `b *> (this <* c)` will be slightly more efficient since it will test b with less depth.

johnynek: actually, to geek out a bit more:
```
(b.void ~ (this ~ c.void)).map { case (_, (a, _)) => a }
```
is probably the most efficient since there is only one map, not two, one for each `<*` and `*>`.

regadas: nice catch! let's do that and save some extra calls

core/shared/src/main/scala/cats/parse/Parser.scala
regadas: gotcha! yeah that makes sense

regadas:
@johnynek good call on trying to add this to the json example. I realised we needed a few more things.

bench/src/main/scala/cats/parse/bench/self.scala
```diff
@@ -47,22 +47,24 @@ object Json {
...
-      (whitespaces0 *> P.repSep(pa, min = 0, sep = listSep) <* whitespaces0)
+      P.repSep(pa, min = 0, sep = listSep).surroundedBy(whitespaces0)
 
-    val list = (P.char('[') *> rep(recurse) <* P.char(']'))
+    val list = rep(recurse).with1
+      .between(P.char('['), P.char(']'))
       .map { vs => JArray.fromSeq(vs) }
 
     val kv: P1[(String, JValue)] =
       justStr ~ ((whitespaces0.with1 ~ P.char(':') ~ whitespaces0) *> recurse)
```
johnynek, line 56: This can use `P.char(':').surroundedBy(whitespace0)`

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1104,6 +1104,24 @@ class ParserTest extends munit.ScalaCheckSuite {
...
+  property("parse between open and close") {
+    forAll(ParserGen.gen1, ParserGen.gen, Arbitrary.arbitrary[String]) { (genP1, genP, str) =>
+      val pa = genP1.fa.between(genP.fa, genP.fa)
+      val pb = genP.fa *> genP1.fa <* genP.fa
+
+      assertEquals(pa.parse(str), pb.parse(str))
+    }
+  }
+
+  property("sorroundedBy consistent with between") {
```
johnynek, line 1116: `surroundedBy` is misspelled.

docs/src/main/paradox/index.md
```diff
@@ -51,15 +51,17 @@ object Json {
...
-      (whitespaces0 *> P.repSep(pa, min = 0, sep = listSep) <* whitespaces0)
+      P.repSep(pa, min = 0, sep = listSep).surroundedBy(whitespaces0)
 
-    val list = (P.char('[') *> rep(recurse) <* P.char(']'))
+    val list = rep(recurse).with1
+      .between(P.char('['), P.char(']'))
       .map { vs => JArray.fromSeq(vs) }
 
     val kv: Parser1[(String, JValue)] =
       justStr ~ ((whitespaces0.with1 ~ P.char(':') ~ whitespaces0) *> recurse)
```
johnynek, line 60: We can use surroundedBy here.

johnynek (commented):
Here are a few nits. If you want to address fine. They core API has no issues.

I really like the addition of with1.between which came up in your use. A good reason to use combinators in our example code!

Thanks for adding these. I think they are common and since we can right-associate it better for the user and avoid `*>` which might not be very beginner friendly, I think it is nice.

bench/src/main/scala/cats/parse/bench/self.scala
regadas: ah yes!

core/shared/src/test/scala/cats/parse/ParserTest.scala
regadas: 🤦

johnynek (approved):

PR #83, opened 2020-11-24
Aggregate bench project

regadas:
With this we propagate `compile` and `fmt` tasks/commands.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,10 @@ jobs:
       - name: Check that workflows are up to date
         run: sbt ++${{ matrix.scala }} githubWorkflowCheck
 
-      - run: sbt ++${{ matrix.scala }} fmtCheck test mimaReportBinaryIssues
+      - run: |
+          sbt ++${{ matrix.scala }} fmtCheck \
+              "++${{ matrix.scala }} test" \
+              "++${{ matrix.scala }} mimaReportBinaryIssues"
 
       - name: Compress target directories
         run: tar cf targets.tar target core/js/target core/jvm/target project/target
--- a/build.sbt
+++ b/build.sbt
@@ -84,7 +84,8 @@ ThisBuild / testFrameworks += new TestFramework("munit.Framework")
 
 lazy val root = project
   .in(file("."))
-  .aggregate(core.jvm, core.js)
+  .settings(scalaVersion := "2.13.3")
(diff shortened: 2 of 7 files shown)


regadas:
hm, need to dig a little bit more as this runs locally? 🤔

regadas:
🙌 alright so I guess this is now doing the expected.

.github/workflows/ci.yml
```diff
@@ -55,7 +55,10 @@ jobs:
       - name: Check that workflows are up to date
         run: sbt ++${{ matrix.scala }} githubWorkflowCheck
 
-      - run: sbt ++${{ matrix.scala }} fmtCheck test mimaReportBinaryIssues
+      - run: |
+          sbt ++${{ matrix.scala }} fmtCheck \
```
johnynek, line 59: why do we need this change? Also, it is interesting there is no quote on the first item.

build.sbt
```diff
@@ -26,7 +26,13 @@ ThisBuild / githubWorkflowPublishTargetBranches := Seq(
 )
 
 ThisBuild / githubWorkflowBuild := Seq(
-  WorkflowStep.Sbt(List("fmtCheck", "test", "mimaReportBinaryIssues"))
+  WorkflowStep.Run(
+    List(
+      """sbt ++${{ matrix.scala }} fmtCheck \
```
johnynek, line 31: I guess this is the difference, note the fmtCheck doesn't have quotes around it... But does it need the `++` at all? The format check does not depend on the scala version does it?

johnynek (commented):
Thanks for dealing with this. I know it can be a frustration to deal with the build.

build.sbt
regadas: > The format check does not depend on the scala version does it?

selecting the scala version on format will allow us to check the format on scala version specific code i.e under`scala-2.1x` folders. Currently we don't have any so we don't really need it. I thought of separating the "checks" into a separate task and run it for one version only but ended up keeping the previous behaviour; can be handy? why not?

The quotes are an interesting case and one of the weirdest behaviours in SBT. When running sbt like this(non interactive) Scala version is not correctly selected on projects when running tasks ... quoting them helps. However, the behavior is not the same on cmds like `fmtCheck` and leads to error when it expands.

Note: from sbt console you don't need the quotes.

johnynek: Thanks for the detailed reply.

johnynek (approved):

PR #82, opened 2020-11-23
Right associate products in unmap

johnynek:
This idea is that in unmap, we can right associate without any additional cost, so why not.

This means if the first item in a product does not match, we can see that quickly. Since unmap is used by void and string, both very common, this should be a meaningful optimization.

Here are benchmarks of this PR:
```
[info] Benchmark                        Mode  Cnt    Score    Error  Units
[info] BarBench.catsParseParse          avgt    5    0.001 ±  0.001  ms/op
[info] BarBench.fastparseParse          avgt    5   ≈ 10⁻⁴           ms/op
[info] BarBench.jawnParse               avgt    5   ≈ 10⁻⁴           ms/op
[info] BarBench.parboiled2Parse         avgt    5   ≈ 10⁻⁴           ms/op
[info] BarBench.parsleyParseCold        avgt    5    0.154 ±  0.004  ms/op
[info] Bla25Bench.catsParseParse        avgt    5   74.026 ±  3.057  ms/op
[info] Bla25Bench.fastparseParse        avgt    5   40.463 ±  0.902  ms/op
[info] Bla25Bench.jawnParse             avgt    5   15.587 ±  2.083  ms/op
[info] Bla25Bench.parboiled2Parse       avgt    5   56.473 ±  4.402  ms/op
[info] Bla25Bench.parsleyParseCold      avgt    5   95.217 ±  4.747  ms/op
[info] CountriesBench.catsParseParse    avgt    5   25.513 ±  1.547  ms/op
[info] CountriesBench.fastparseParse    avgt    5   13.602 ±  0.282  ms/op
[info] CountriesBench.jawnParse         avgt    5    3.411 ±  0.135  ms/op
[info] CountriesBench.parboiled2Parse   avgt    5   11.239 ±  0.658  ms/op
[info] CountriesBench.parsleyParseCold  avgt    5   34.266 ±  7.549  ms/op
[info] Qux2Bench.catsParseParse         avgt    5   20.933 ±  0.641  ms/op
[info] Qux2Bench.fastparseParse         avgt    5   16.323 ±  0.530  ms/op
[info] Qux2Bench.jawnParse              avgt    5    4.338 ±  0.135  ms/op
[info] Qux2Bench.parboiled2Parse        avgt    5   16.619 ±  0.591  ms/op
[info] Qux2Bench.parsleyParseCold       avgt    5   31.967 ±  1.007  ms/op
[info] Ugh10kBench.catsParseParse       avgt    5  175.986 ±  7.386  ms/op
[...]

PR #86, opened 2020-11-25
Increase test coverage a bit

johnynek:
This increases the test coverage of the recent #79 PR.

Also, a minor improvement of unmap for right association #82 - we use Void which acts as a signal that something is already unmapped.

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -1352,8 +1352,10 @@ object Parser extends ParserInstances {
           unmap(p1) match {
             case Prod(p11, p12) =>
               // right associate so
-              // we can find check matches a bit faster
-              Prod(p11, unmap(Prod(p12, p2)))
+              // we can check matches a bit faster
+              // note: p12 is already unmapped, so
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1417,12 +1421,15 @@ object Parser extends ParserInstances {
           unmap(p1) match {
             case Prod(p11, p12) =>
               // right associate so
-              // we can find check matches a bit faster
-              Prod1(p11, unmap(Parser.product(p12, p2)))
+              // we can check matches a bit faster
+              // note: p12 is already unmapped, so
+              // we wrap with Void to prevent n^2 cost
+              Prod1(p11, unmap(Parser.product(p12.void, p2)))
```
johnynek, line 1427: we use `.void` here because if `p12` is a parser1 but `p2` isn't we need to make sure to keep the `p12.void` to be a Parser1 and .void does this. Otherwise we would need to test here.

regadas (approved):
LGTM!

Thanks for adding the extra tests, that's my fault.

johnynek:
no problem! Thanks for doing the work!

PR #89, opened 2020-11-26
Update scala versions

regadas:
* 2.13.4
* 3.0.0-M2


mpilquist:
Just saw this after opening #93 - I think we should adopt the philosophy of publishing for last 2 dotty releases until 3.0.0 final. I know http4s will be on 3.0.0-M1 for a while as the various projects build for M2.

regadas:
no problem 👍 since you are doing in #93 + some cleanup I don't mind closing this.

PR #93, opened 2020-11-26
Add Scala 3.0.0-M2 support and cleanup build

mpilquist:


build.sbt
```diff
@@ -53,22 +53,6 @@ ThisBuild / githubWorkflowAddedJobs ++= Seq(
   )
 )
 
-ThisBuild / githubWorkflowEnv ++= Map(
```
johnynek, line 56: Why is this no longer needed?

johnynek (commented):
I don't really understand why that block can be removed from the build.

But I always appreciate your help.

build.sbt
mpilquist: I convinced Daniel to take all the boilerplate needed to publish to Sonatype and put it in sbt-spiewak. We get all the removed config via the new `SonatypeCiRelease` plugin enabled on the root project. Documented here: https://github.com/djspiewak/sbt-spiewak#ci-release

mpilquist:
By the way, this also moves us to cats 2.3.0 final, which is cross built for scala 3.0.0-M1 & 3.0.0-M2 -- no more _2.13 suffix clashes. Scala.js builds for Scala 3 M1/M2 are also enabled now.

johnynek:
Okay cool! That plug-in looks great. Kudos to you and Daniel.

The CI seemed to fail on fmt checking. I didn't investigate more yet.

build.sbt
```diff
@@ -90,14 +74,11 @@ ThisBuild / testFrameworks += new TestFramework("munit.Framework")
 
 lazy val root = project
   .in(file("."))
-  .settings(scalaVersion := "2.13.3")
```
regadas, line 93: I believe build is failing because this is settings required to cross-compile correctly with the aggregated projects.

build.sbt
```diff
@@ -126,14 +126,20 @@ lazy val bench = project
...
+        Seq(
+          fastParse,
+          parsley,
+          jawnAst,
+          parboiled,
+          attoCore
+        )
+      else Nil
+    },
     githubWorkflowArtifactUpload := false
```
regadas, line 143: Why not just adding back the `scalaVersion` setting? This seems more complicated and doesn't work well with SBT cross-compile. this yields 
```
sbt:root> ++3.0.0-M2 -v
[info] Setting Scala version to 3.0.0-M2 on 4 projects.
[info] Switching Scala version on:
[info]     docs (3.0.0-M2, 3.0.0-M1, 2.12.12, 2.13.4)
[info]     coreJS (3.0.0-M2, 3.0.0-M1, 2.12.12, 2.13.4)
[info]     coreJVM (3.0.0-M2, 3.0.0-M1, 2.12.12, 2.13.4)
[info]     bench (3.0.0-M2, 3.0.0-M1, 2.12.12, 2.13.4)
[info] Excluding projects:
[info]   * root (2.13.4)
```

and with the setting 

```
[info] Setting Scala version to 3.0.0-M2 on 3 projects.
[info] Switching Scala version on:
[info]     docs (3.0.0-M2, 3.0.0-M1, 2.12.12, 2.13.4)
[info]     coreJS (3.0.0-M2, 3.0.0-M1, 2.12.12, 2.13.4)
[info]     coreJVM (3.0.0-M2, 3.0.0-M1, 2.12.12, 2.13.4)
[info] Excluding projects:
[info]   * root (2.13.4)
[info]     bench (2.12.12, 2.13.4)
[info] Reapplying settings...
```

mpilquist: It’s not that simple... it results is cross suffix version clashes, when then need exclusions, which then cause compilation issues

regadas: hm main was compiling fine. Have you tried it?

mpilquist: Yep I did - it’s the cats-core 2.3.0 dependency with its new 3.0.0-MC cross builds

regadas: 🤔 Hm maybe SBT acting up but cats-core 2.3.0 is also on main.

mpilquist: Yeah, but on main, we're still using the _2.13 version of cats-core in the Dotty builds via `withDottyCompat` (which is hidden in the `.settings(dottyLibrarySettings)`.

regadas: Hm I see but all of this happens because `bench` is being targeted to compile with `3.0.0-MC`. As an experiment I took this PR up until https://github.com/regadas/cats-parse/commit/2b2d83d49be56fff9f9ba3348a97b529f45fdc87 and applied https://github.com/regadas/cats-parse/commit/fc680ee8c9c08c4bb8e578f7e9aa3e90c6ec6e5c and it seemed to work https://github.com/regadas/cats-parse/actions/runs/385859531 so I'm curious why it didn't work for you.

Sorry for geeking out, I'm always puzzled with these SBT quirks and I try to keep things as simple as possible so we don't ask "why we have this in the future"

mpilquist: I don't know but if you take this PR and apply this diff:

```
diff --git a/build.sbt b/build.sbt
index 571720d..a757df7 100644
--- a/build.sbt
+++ b/build.sbt
@@ -76,6 +76,7 @@ lazy val root = project
   .in(file("."))
   .aggregate(core.jvm, core.js, bench)
   .enablePlugins(NoPublishPlugin, SonatypeCiRelease)
+  .settings(scalaVersion := "2.13.4")
 
 lazy val docs = project
   .enablePlugins(ParadoxPlugin, MdocPlugin, NoPublishPlugin)
@@ -126,20 +127,14 @@ lazy val bench = project
   .settings(
     name := "bench",
     coverageEnabled := false,
-    Compile / sources := {
-      if (scalaVersion.value.startsWith("2.")) (Compile / sources).value else Nil
-    },
-    libraryDependencies ++= {
-      if (scalaVersion.value.startsWith("2."))
-        Seq(
-          fastParse,
-          parsley,
-          jawnAst,
-          parboiled,
-          attoCore
-        )
-      else Nil
-    },
+    libraryDependencies ++=
+      Seq(
+        fastParse,
+        parsley,
+        jawnAst,
+        parboiled,
+        attoCore
+      ),
     githubWorkflowArtifactUpload := false
   )
   .dependsOn(coreJVM)
```

We get this error when running `+test` from root:

```
[error] stack trace is suppressed; run last bench / update for the full output
[error] (bench / update) sbt.librarymanagement.ResolveException: Error downloading org.tpolecat:atto-core_3.0.0-M1:0.8.0
```

regadas: ahh I see ... ok so in your diff you're missing `crossScalaVersions`:

```
    crossScalaVersions := (ThisBuild / crossScalaVersions).value.filter(_.startsWith("2.")),
```

like it is on main:

https://github.com/typelevel/cats-parse/blob/e1a9ebfd87c4035b6314e153cd2d3c34e12d0365/build.sbt#L150-L160

johnynek (approved):
Thanks @mpilquist !

I really appreciate the build help. It can be pretty finicky to get these highly cross built things working.

PR #94, opened 2020-11-26
Ignore bench from codecov

regadas:
Easy way to increase codecov

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,2 @@
+ignore:
+    - "bench" # ignore bench project and all its contents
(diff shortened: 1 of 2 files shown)


codecov.yml
```diff
@@ -0,0 +1,2 @@
+ignore:
```
regadas, line 1: actually I just remembered we can use 

smth like `coverageExcludedPackages := "com\\.project\\.core\\.exclude.*"`

regadas: `coverageEnabled := false,` does the job.

johnynek (commented):
Thanks!

PR #96, opened 2020-11-26
Simplify bench project Scala version exclusions

mpilquist:
@regadas This is working for me after all - not sure why it didn't work initially, perhaps I lost track while trying different things.


regadas (approved):
@mpilquist awesome! Thank you!

PR #39, opened 2020-11-04
Site [in progress]

Slakah:
Fixes #7

TODO:
- [ ] Logo
- [ ] Header
- [ ] Footer
- [ ] Home page
- [ ] Publish site GitHub action
- [ ] Add [medium blog post](https://posco.medium.com/designing-a-parsing-library-in-scala-d5076de52536) to cats-parse blog
- [ ] Site versioning (optional) could potentially lift the scripts from courier see: [here](https://github.com/coursier/coursier/blob/master/website.sc)

PR #98, opened 2020-11-27
Add material theme to paradox

regadas:
I know #39 is still being worked on but since we have already paradox I thought we could put it to some use in the meantime? wdyt is it worth it?

page


johnynek:
seems pretty nice to me. Seems pretty easy to merge and change later.

That said, is the site being published?

regadas:
Hi @johnynek, sorry for the late reply! It's not being pushed; I intend to make it so. Just wanted to see if this would be nice to have first. 😄

I'll update this later. Thanks!

regadas:
Hi @johnynek took me longer than expected to follow up on this. Added a step to CI to make the site automatically published on release.

project/plugins.sbt
```diff
@@ -9,3 +9,6 @@ addSbtPlugin("com.codecommit" % "sbt-spiewak-sonatype" % "0.19.1")
 addSbtPlugin("com.codecommit" % "sbt-github-actions" % "0.9.5")
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.2")
 addSbtPlugin("org.scalameta" % "sbt-mdoc" % "2.2.13")
+addSbtPlugin("io.github.jonas" % "sbt-paradox-material-theme" % "0.6.0")
+addSbtPlugin("com.typesafe.sbt" % "sbt-site" % "1.4.1")
+addSbtPlugin("com.typesafe.sbt" % "sbt-ghpages" % "0.6.3")
```
regadas, line 14: These allows us to easily do adhoc site pushes through `docs/ghpagesPushSite`

johnynek:
This is great! I personally think the win of using an all scala/sbt solution is worth it in terms of simplicity of the CI and build.

johnynek (approved):

rossabaker:
:tada: We use this on http4s-jdk-http-client, and I would love to see it spread.  The multi-version support is lacking, but other than that, it's a solid tool and an attractive, functional theme.  And no futzing with Jekyll or Hugo is a big win.

johnynek:
Related to #7

PR #99, opened 2020-11-30
Fix return type on repAs1

johnynek:
This was an oversight: the repAs1 extension method hid that it returns a Parser1

The work around is to directly call the repAs1 method on the Parser companion which does return the correct type.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ addCommandAlias("fmtCheck", "; scalafmtCheckAll; scalafmtSbtCheck")
 
 addCommandAlias("prePR", "; githubWorkflowGenerate ; +fmt; bench/compile; +test")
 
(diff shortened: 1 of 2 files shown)


regadas (approved):
👍 indeed this makes sense! but now we have MiMa complaining.

johnynek:
@mpilquist since mima stuff seems to be pretty automatic, how is this kind of PR generally merged in such cases? Do we add the exclusions, then publish 0.2, then remove the exclusions?

rossabaker:
If you bump the `baseVersion` to 0.2, I don't think it will check against 0.1 anymore.

johnynek:
actually, all the classes in Parser are sealed or final, so I think exclusions would be fine, but `0.1.0` is a low version anyway, so why not go to `0.2.0` when we publish next (I think after adding #101 ).

PR #101, opened 2020-12-06
filter, based on selective

rossabaker:
Riffing on the ideas of [Staged Selective Parser Combinators](https://mpickering.github.io/papers/parsley-icfp.pdf).

Use case: in http4s, we need to parse various date headers into a 7-tuple of (weekday, year, month, day, hour, minute, second), and then verify that the date is valid (e.g., no 31st of February).  Currently we have to flatMap it, and when parsing, we _don't_ want to flatMap that shit.

Implemented on `branch`.  We should have the full power of Selective.  The primitive operation could also be `select`, which is the abstract function on cats-selective.

Should be able to make a version that fails with a particular message, swapping that in for "empty".

There may be optimizations to sprinkle throughout the code.  Fusing to `Impl.Map` seems possible, for instance.

I'm super sleepy and just following types.  This might all be nonsense.

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -1745,8 +1745,7 @@ object Parser extends ParserInstances {
       else null.asInstanceOf[B]
     }
 
-    case class Select[A, B](parser: Parser[Either[A, B]], fn: Parser[A => B])
-        extends Parser[B] {
+    case class Select[A, B](parser: Parser[Either[A, B]], fn: Parser[A => B]) extends Parser[B] {
       override def parseMut(state: State): B = Impl.select[A, B](parser, fn, state)
     }
 
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -189,6 +189,9 @@ sealed abstract class Parser[+A] {
   def map[B](fn: A => B): Parser[B] =
     Parser.map(this)(fn)
 
+  def filter(fn: A => Boolean): Parser[A] =
+    Parser.select(this.map(a => Either.cond(fn(a), a, ())))(Parser.Fail)
+
   /** Dynamically construct the next parser based on the previously
     * parsed value.
     *
@@ -911,6 +914,9 @@ object Parser extends ParserInstances {
       case _ => Impl.Map1(p, fn)
     }
 
+  def select[A, B](p: Parser[Either[A, B]])(fn: Parser[A => B]): Parser[B] =
+    Impl.Select(p, fn)
+
   /** Standard monadic flatMap
     *  Avoid this function if possible. If you can
     *  instead use product, ~, *>, or <* use that.
@@ -1331,6 +1337,9 @@ object Parser extends ParserInstances {
         case Map(p, _) =>
           // we discard any allocations done by fn
           unmap(p)
+        case Select(p, _) =>
+          // we discard any allocations done by fn
+          unmap(p)
         case StringP(s) =>
           // StringP is added privately, and only after unmap
           s
@@ -1720,6 +1729,22 @@ object Parser extends ParserInstances {
 
     case class Map1[A, B](parser: Parser1[A], fn: A => B) extends Parser1[B] {
       override def parseMut(state: State): B = Impl.map(parser, fn, state)
+    }
+
+    final def select[A, B](
+        parser: Parser[Either[A, B]],
+        fn: Parser[A => B],
+        state: State
+    ): B = {
+      val either = parser.parseMut(state)
+      if ((state.error eq null) && state.capture)
+        either.valueOr(a => fn.map(_(a)).parseMut(state))
+      else null.asInstanceOf[B]
+    }
+
+    case class Select[A, B](parser: Parser[Either[A, B]], fn: Parser[A => B])
+        extends Parser[B] {
+      override def parseMut(state: State): B = Impl.select[A, B](parser, fn, state)
     }
 
     final def flatMap[A, B](parser: Parser[A], fn: A => Parser[B], state: State): B = {
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -1338,8 +1338,7 @@ object Parser extends ParserInstances {
           // we discard any allocations done by fn
           unmap(p)
         case Select(p, fn) =>
-          // we discard any allocations done by fn
-          Select(p, unmap(fn).map(Function.const))
+          Select(p, unmap(fn).asInstanceOf[Parser[Any => Any]])
         case StringP(s) =>
           // StringP is added privately, and only after unmap
           s
@@ -1742,7 +1741,13 @@ object Parser extends ParserInstances {
       state.capture = cap
       if (state.error eq null)
         either match {
-          case Left(a) => fn.map(_(a)).parseMut(state)
+          case Left(a) =>
+            if (state.capture) {
+              fn.map(_(a)).parseMut(state)
+            } else {
+              fn.parseMut(state)
+              null.asInstanceOf[B]
+            }
           case Right(b) => b
         }
       else null.asInstanceOf[B]
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -1337,9 +1337,9 @@ object Parser extends ParserInstances {
         case Map(p, _) =>
           // we discard any allocations done by fn
           unmap(p)
-        case Select(p, _) =>
+        case Select(p, fn) =>
           // we discard any allocations done by fn
-          unmap(p)
+          Select(p, unmap(fn).map(Function.const))
         case StringP(s) =>
           // StringP is added privately, and only after unmap
           s
@@ -1736,8 +1736,11 @@ object Parser extends ParserInstances {
         fn: Parser[A => B],
         state: State
     ): B = {
+      val cap = state.capture
+      state.capture = true
       val either = parser.parseMut(state)
-      if ((state.error eq null) && state.capture)
+      state.capture = cap
+      if (state.error eq null)
         either match {
           case Left(a) => fn.map(_(a)).parseMut(state)
           case Right(b) => b
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -1738,7 +1738,10 @@ object Parser extends ParserInstances {
     ): B = {
       val either = parser.parseMut(state)
       if ((state.error eq null) && state.capture)
-        either.valueOr(a => fn.map(_(a)).parseMut(state))
+        either match {
+          case Left(a) => fn.map(_(a)).parseMut(state)
+          case Right(b) => b
+        }
       else null.asInstanceOf[B]
     }
 
--- a/core/shared/src/test/scala/cats/parse/ParserTest.scala
+++ b/core/shared/src/test/scala/cats/parse/ParserTest.scala
@@ -1561,4 +1561,18 @@ class ParserTest extends munit.ScalaCheckSuite {
     }
   }
 
+  property("p.filter(_ => true) == p") {
+    forAll(ParserGen.gen, Arbitrary.arbitrary[String]) { (genP, str) =>
+      val res0 = genP.fa.filter(_ => true).parse(str)
+      val res1 = genP.fa.parse(str)
+      assertEquals(res0, res1)
+    }
+  }
+
+  property("p.filter(_ => false) fails") {
+    forAll(ParserGen.gen, Arbitrary.arbitrary[String]) { (genP, str) =>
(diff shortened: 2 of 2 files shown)


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1599,6 +1607,23 @@ object Parser extends ParserInstances {
...
 
+    final def branch[A, B, C](
+        parser: Parser[Either[A, B]],
+        l: Parser[A => C],
+        r: Parser[B => C],
+        state: State
+    ): C = {
+      val either = parser.parseMut(state)
+      if ((state.error eq null) && state.capture)
+        either.fold(a => l.ap(Parser.pure(a)), b => r.ap(Parser.pure(b))).parseMut(state)
```
johnynek: Let's not use fold. Pattern matching avoids allocating a closure and is generally faster, and this could appear in a tight loop inside a rep.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1499,4 +1499,18 @@ class ParserTest extends munit.ScalaCheckSuite {
...
+      assertEquals(res0, res1)
+    }
+  }
+
+  property("p.filter(_ => false) fails") {
+    forAll(ParserGen.gen, Arbitrary.arbitrary[String]) { (genP, str) =>
+      val res = genP.fa.filter(_ => false).parse(str)
+      assert(res.isLeft)
+    }
+  }
```
johnynek, line 1577: Can we write a law on branch directly (see the laws I have on how ~ composes). I don't think what we have now is exactly right but since filter is weaker is might not detect it.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1599,6 +1607,23 @@ object Parser extends ParserInstances {
...
+      else null.asInstanceOf[C]
```
johnynek: I think this isn't right. I think if we are not capturing we still need to run the result because it may advance the offset in the parser.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1499,4 +1499,18 @@ class ParserTest extends munit.ScalaCheckSuite {
     }
   }
 
+  property("p.filter(_ => true) == p") {
```
johnynek, line 1564: Since branch is a new fundamental operation I think we need to add it to the generators. That will test to make sure none of the past properties are broken by it's presence.

johnynek (changes requested):
This is really cool. I think we should add it. 

One note: users can implement filter, with less performance, using flatMap in the mean time (I guess any time you have Monad + Alternative this is true).

I had a couple concerns that I noted.

rossabaker:
Ugh.  I didn't think anybody had looked at it yet, and just force-pushed it based on `select` and fixing some 3am mistakes. But I think a lot of your feedback is still valid.

I'd also like to throw in an example and perhaps a bench to show how it compares to the monadic approach.

core/shared/src/main/scala/cats/parse/Parser.scala
rossabaker: :+1: I'll inline the `valueOr` that replaced it.

johnynek: Cool. And thinking more, I don't think we want to use ap. We should run parseMut on the currect side then apply the function ourselves.

johnynek:
Dude I check my GitHub notifications religiously. Well really more than most people do religious things.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1722,6 +1731,22 @@ object Parser extends ParserInstances {
       override def parseMut(state: State): B = Impl.map(parser, fn, state)
     }
 
+    final def select[A, B](
+        parser: Parser[Either[A, B]],
+        fn: Parser[A => B],
+        state: State
+    ): B = {
+      val either = parser.parseMut(state)
```
johnynek, line 1739: Actually, look at flatMap. I think you have to set capture to true here unconditionally since otherwise we don't know what to do next.

After we capture the first, we can set capture to false on the second part since we can just avoid applying the function.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -911,6 +914,9 @@ object Parser extends ParserInstances {
       case _ => Impl.Map1(p, fn)
     }
 
+  def select[A, B](p: Parser[Either[A, B]])(fn: Parser[A => B]): Parser[B] =
```
johnynek, line 917: For bonus points we can add select10 that takes a Parser1 then a Parser and returns a Parser1. Similarly a select01 that does the Parser1 second.

If we had that, we can override filter on Parser1 to return a Parser1

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1561,4 +1561,18 @@ class ParserTest extends munit.ScalaCheckSuite {
...
+    forAll(ParserGen.gen, Arbitrary.arbitrary[String]) { (genP, str) =>
+      val res0 = genP.fa.filter(_ => true).parse(str)
+      val res1 = genP.fa.parse(str)
+      assertEquals(res0, res1)
+    }
+  }
+
```
johnynek, line 1571: Another interesting pair of laws: if we pa.ap(pb) should be the same as select(pa.map(Left), pb, fail)

Similarly on the right hand side.

core/shared/src/main/scala/cats/parse/Parser.scala
rossabaker: Yep, already rid of that locally. I was translating a bit too literally from Haskell last night.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1331,6 +1337,9 @@ object Parser extends ParserInstances {
         case Map(p, _) =>
           // we discard any allocations done by fn
           unmap(p)
+        case Select(p, _) =>
+          // we discard any allocations done by fn
+          unmap(p)
```
johnynek, line 1342: I think you need to do Select(a, b) => Select(a, unmap(b))

Since in select, after we optionally parse the right in a void we could not do anything more, but we do need the result on the left even in a void.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -244,6 +244,25 @@ object ParserGen {
...
   }
 
+  def selected(ga: Gen[GenT[Parser]]): Gen[GenT[Parser]] =
+    Gen.zip(ga, pures).flatMap { case (parser, genRes) =>
+      val gpab = Gen.either(parser.fa, genRes.fa.map(Parser.pure)).map {
+        _ match {
+          case Left(pa) => pa.map(Left(_))
+          case Right(pb) => pb.map(Right(_))
+        }
+      }
```
rossabaker, line 254: Feels like there is a $10 function I'm overlooking here.  I think Haskell calls it `uncozip`.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -244,6 +244,25 @@ object ParserGen {
     val fn: A => F[B]
   }
 
+  def selected(ga: Gen[GenT[Parser]]): Gen[GenT[Parser]] =
+    Gen.zip(ga, pures).flatMap { case (parser, genRes) =>
+      val gpab = Gen.either(parser.fa, genRes.fa.map(Parser.pure)).map {
```
rossabaker, line 249: We need a `B` that can be the right of `gpab` and also the return type of the function in `gfn`.  This compiles, but the `Parser.pure` feels shoddy.  I'll try Level 2 Type Tetris later.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -244,6 +244,25 @@ object ParserGen {
...
+        _ match {
+          case Left(pa) => pa.map(Left(_))
+          case Right(pb) => pb.map(Right(_))
+        }
+      }
+
+      val gfn: Gen[Parser[parser.A => genRes.A]] =
+        Gen.function1(genRes.fa)(parser.cogen).map(Parser.pure)
```
rossabaker, line 257: Another dubious `Parser.pure`.

core/shared/src/test/scala/cats/parse/ParserTest.scala
johnynek: Yeah my claim is with a pure here the laws won't pass without more changes.

core/shared/src/test/scala/cats/parse/ParserTest.scala
johnynek: Why the match? Why not just do the case at the top level?

rossabaker: Because I've been locked down with these kids since March and am getting a little duller by the hour.

core/shared/src/test/scala/cats/parse/ParserTest.scala
rossabaker: Okay, Type Tetris Two looks more like the tailRecM case, and got us some failing tests.  I think now we can see if your other suggestions green it back up.

core/shared/src/main/scala/cats/parse/Parser.scala
rossabaker: The second parameter has to be a function, so I had to immediately and trivially map the unmapped value.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1722,6 +1731,27 @@ object Parser extends ParserInstances {
...
+        fn: Parser[A => B],
+        state: State
+    ): B = {
+      val cap = state.capture
+      state.capture = true
+      val either = parser.parseMut(state)
+      state.capture = cap
+      if (state.error eq null)
+        either match {
+          case Left(a) => fn.map(_(a)).parseMut(state)
```
johnynek, line 1745: I think you can do fun.parseMut(state) and in the capture case actually apply the function. In the non capture case you can just return null.

I think this also solves the unmap requirement of putting in const.

core/shared/src/main/scala/cats/parse/Parser.scala
johnynek: Actually with this formulation I think we only need select and select1. In select1 the first argument is a Parser1.

core/shared/src/main/scala/cats/parse/Parser.scala
rossabaker: The `.asInstanceOf` in the `unmap` makes me queasy, but it makes the existing tests pass.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1588,6 +1588,22 @@ class ParserTest extends munit.ScalaCheckSuite {
     }
   }
 
+  property("select(pa.map(Left(_)))(pf) == pf.ap(pa)") {
+    forAll(ParserGen.gen, ParserGen.gen) { (genP, genRes) =>
+      val pa = genP.fa
+      val pf = null: Parser[genP.A => genRes.A]
```
rossabaker, line 1594: I'm struggling to produce this parser.  The trick used in the generator didn't quite translate.

johnynek:
I made a PR as a suggestion to this one:

https://github.com/rossabaker/cats-parse/pull/1

It gets everything green, but changes somewhat the internal representation of Select (to work more cleanly with unmap).

johnynek (approved):
What do you think about merging this and publishing 0.2.0?

I think we need to remove the "Draft" label.

rossabaker:
I'm good with that.  I'd still like to figure out how to write that Left law, and I'm also interested in a benchmark that shows the advantage (or not) over flatMap.  But none of that needs to hold back progress.

mpilquist:
Bump `ThisBuild / baseVersion` before doing so -- not sure that failing to do so will cause a problem but should do it anyway.

johnynek:
Actually I think adding an override for ifM using select:

https://github.com/typelevel/cats/blob/8e8be51187385dde78321e7e0d9fd892801afacb/core/src/main/scala/cats/FlatMap.scala#L123

We should be able but I don't see it immediately. I assume we use select twice.

rossabaker:
`baseVersion` is already bumped.

Devil's advocacy: a risk in merging this before a benchmark is that the gains may be only theoretical and not justify the new primitive.  I definitely want `filter`, but there's also a monadic way to express it.

That said, I think it _is_ better, and even if it's not, it's still 0.x.

johnynek:
I bet ifM will be a lot faster using select (or a similar internal node).

rossabaker:
cats-selective and its Haskell forbears call it `ifS`, and we'd get it for free with a cats-selective integration.

If this proves successful, I'd be interested in renewing the push to get Selective into Cats proper.  With the right encoding, I think it could even be accepted into 2.x.

regadas:
This is really cool! 🎉

PR #103, opened 2020-12-06
Add LocationMap to help users show error locations

johnynek:
This adds some code from bosatsu which converts offsets in the string into lines and columns.

This is intended for users to convert offsets in errors into snippets to display to the user (by combining toLineCol with getLine).

@rossabaker I'd love your review here.

In bosatsu, I have a bit more, a function to show the context and to point to the offset, but it uses paiges Doc, and I don't want to add that dependency at this time.

I thought a bit about how to abstract that out, but it isn't trivial. I figured this part is at least part of what we want.

### DIFF
--- a/core/shared/src/main/scala/cats/parse/LocationMap.scala
+++ b/core/shared/src/main/scala/cats/parse/LocationMap.scala
@@ -1,0 +1,93 @@
+/*
+ * Copyright (c) 2020 Typelevel
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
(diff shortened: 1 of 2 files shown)


rossabaker (approved):
This is great.  We're just calling toString on the Left in http4s right now, and this improvements built on this will be great.

I wouldn't object to naming the tuple `Pos` or something.

Another property I think is true, though uninteresting.  I think your coverage is already outstanding:

* if `x > y && toLineCol(x).isDefined`, then `toLineCol(x) > toLineCol(y)`

We'll have a lot of `\r\n`, and I think this will do the intuitive thing on those.  We're not treating `\r` as a line delimiter, and I think the semantics get much uglier supporting it as such for little gain.

PR #107, opened 2020-12-09
Add a simple law for select

johnynek:
This is a law that shows the when first argument of the select is Right, we don't look at the second parser.

### DIFF
--- a/core/shared/src/test/scala/cats/parse/ParserTest.scala
+++ b/core/shared/src/test/scala/cats/parse/ParserTest.scala
@@ -1647,4 +1647,28 @@ class ParserTest extends munit.ScalaCheckSuite {
       assert(res.isLeft)
     }
   }
+
+  property("select on pure values works as expected") {
+    forAll { (left: Option[Either[Int, String]], right: Option[Int => String], str: String) =>
+      val pleft = left match {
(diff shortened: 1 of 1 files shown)


rossabaker (approved):
This looks consistent with the Haskell law that `select = selectM` if the selective is also a monad:

```haskell
selectM :: Monad f => f (Either a b) -> f (a -> b) -> f b
selectM mx mf = do
    x <- mx
    case x of
        Left  a -> fmap ($a) mf
        Right b -> pure b
```

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1647,4 +1647,28 @@ class ParserTest extends munit.ScalaCheckSuite {
       assert(res.isLeft)
     }
   }
+
+  property("select on pure values works as expected") {
```
ritschwumm, line 1651: somehow the comment on the PR "This is a law that shows that when first argument of the select is Right, we don't look at the second parser." was easier to understand to me than either "works as expected" or the code below.

johnynek: PR for such a change would be welcomed!

PR #108, opened 2020-12-09
Implement FunctorFilter and add some more laws

johnynek:
This does a few things:

1. add an arbitrary for Parser that just maps a totally general one into the right type. This should make those parser internally complex, but have the right type for some simple law writing
2. add a few new laws that are easier to write, or that were overlooked in the past
3. add FunctorFilter methods which are related to filter, but also allows a transformation at the same time, e.g. convert a string to an Int32: usually it is easy to see if a string contains a big integer, but only parsing valid int32 makes a very complex parser and potentially bad error messages (although, using a failWith to explain is probably better here, but that pushes you out of select and into flatMap territory).

what do you think @rossabaker

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -189,6 +189,24 @@ sealed abstract class Parser[+A] {
   def map[B](fn: A => B): Parser[B] =
     Parser.map(this)(fn)
 
+  /** Transform parsed values using the given function, or fail on None
+    *
+    * When the function return None, this parser fails
+    * This is implemented with select, which makes it more efficient
+    * than using flatMap
+    */
+  def mapOrFail[B](fn: A => Option[B]): Parser[B] = {
+    val leftUnit = Left(())
+
+    val first = map { a =>
+      fn(a) match {
+        case Some(b) => Right(b)
+        case None => leftUnit
+      }
+    }
+    Parser.select(first)(Parser.Fail)
+  }
+
   /** If the predicate is not true, fail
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -189,6 +189,24 @@ sealed abstract class Parser[+A] {
   def map[B](fn: A => B): Parser[B] =
     Parser.map(this)(fn)
 
+  /** Transform parsed values using the given function, or fail on None
+    *
+    * When the function return None, this parser fails
+    * This is implemented with select, which makes it more efficient
+    * than using flatMap
+    */
+  def mapOrFail[B](fn: A => Option[B]): Parser[B] = {
```
rossabaker, line 198: This looks like `FunctorFilter#mapFilter`.  Is that a better name?

rossabaker (commented):
Yeah, those arbitraries make it much easier to write the laws I was struggling with.

core/shared/src/main/scala/cats/parse/Parser.scala
regadas: I was thinking the same. Could we add a `def collect[B](fn: PartialFunction[A, B]): Parser[B]` as well?

johnynek: great review folks, I added a FunctorFilter implementation and used these names.

Please take a look. I'd like to publish 0.2.0 after merging this unless you disagree.

regadas (approved):
Awesome!

PR #110, opened 2020-12-11
Add `stringIn1` parser

satabin:
This parser builds a radix tree from non empty string alternatives.
Radix trees are interesting as they allow for efficient parsing of
string alternatives (e.g. enum names) without backtracking and require
only 1 character lookahead.

String alternatives are first sorted and then grouped by common non
empty prefix. Grouped string are then recursively grouped with the same
process. In the resulting tree, each group of substrings start with a
different character, which can be used to decide which alternative
branch to take.

This operator is inspired by fastparse `stringIn` one, and requires all
alternatives to be non empty.


satabin:
Is it ok to depend on `scala-collection-compat` to solve the cross-compilation problem with 2.12?

johnynek (commented):
Thanks for the PR! This is something I have wanted to get to.

I'd like to make two suggestions:

1. Can we remove the stringIn1 API? It seems to me this can be an internal detail of how to combine Str parsers inside of oneOf much as we internally combine CharIn parsers.
2. If you feel we should keep the stringIn API, I'm concerned that we don't seem to be deduplicating it anywhere. I'd rather accept Iterable, then make a distinct set and the if it has only one item just return Str. Also if we keep the API it should have a comment in particular describing the backtracking behavior of the method.
4. Since this is a performance improvement, can you add a benchmark? Users now have two ways to do this: oneOf + string1 or stringIn1, or if we don't want to expose the API internally we still need to choose. We need to show it is faster to use the tree. I suspect it will be in some cases, but maybe not all cases especially when there are few and distinct key words. But a benchmark will help us see. See the current benchmark directory for examples.
5. I have some suggestions on improving the performance of the matching loop but I'll wait till we have benchmarks so we can see if it is working.

Thanks for sending the PR.

core/shared/src/main/scala/cats/parse/RadixNode.scala
```diff
@@ -0,0 +1,97 @@
...
+      }
+    NonEmptyList.fromList(strings.filter(_.nonEmpty)) match {
+      case Some(nonEmpty) =>
+        val grouped =
+          groupByNonEmptyPrefix(
+            nonEmpty.tail,
+            nonEmpty.head,
+            NonEmptyList.one(strings.head),
+            Map.empty
+          ).view.mapValues(fromSortedStrings(_))
```
johnynek, line 67: You can use .map { case (k, v) => (k, fromSortedStrings(v) }

to get this to work on 2.12 and 2.13 without using any compat code.

satabin: This is true, I just went for `mapValues` as I tend to prefer this, but this the solves my cross-compilation problem.

satabin:
> Thanks for the PR! This is something I have wanted to get to.
> 
> I'd like to make two suggestions:
> 
>     1. Can we remove the stringIn1 API? It seems to me this can be an internal detail of how to combine Str parsers inside of oneOf much as we internally combine CharIn parsers.

The current implementation heavily relies of the fact that it takes strings and not parsers. I can see if I can internalize it in `oneOf1` but it seems a bit different in semantics to me currently: there is no need for `.backtrack` when several alternatives share a common prefix. I fear that working below `oneOf` will hide this from us, but I can give it a try.

> 
>     2. If you feel we should keep the stringIn API, I'm concerned that we don't seem to be deduplicating it anywhere. I'd rather accept Iterable, then make a distinct set and the if it has only one item just return Str. Also if we keep the API it should have a comment in particular describing the backtracking behavior of the method.

Deduplication happens implicitly when building the tree, but I can make it explicit first, before I sort. I modeled the pre-conditio after the one from `oneOf` but I can make it smarter to return a `Str` in case there is only one in the end.

I will document the (lack of) backtracking in the API

> 
>     3. Since this is a performance improvement, can you add a benchmark? Users now have two ways to do this: oneOf + string1 or stringIn1, or if we don't want to expose the API internally we still need to choose. We need to show it is faster to use the tree. I suspect it will be in some cases, but maybe not all cases especially when there are few and distinct key words. But a benchmark will help us see. See the current benchmark directory for examples.

I will add those.

> 
>     4. I have some suggestions on improving the performance of the matching loop but I'll wait till we have benchmarks so we can see if it is working.
> 

Sure, let's see how it can be further improved.

> 
> Thanks for sending the PR.

satabin:
Hello @johnynek,

I reworked this PR and made the following changes:
 - added a benchmark for the `stringIn1` implementation
 - made the operator smarter in case there is only 0 or 1 alternatives
 - rewrote the radix tree implementation to use an array map, as the benchmark showed the `Map` was a big bottleneck (which I was supposing, but the benchmark showed it clearly)

I still didn't merge it with `oneOf1` since the semantics of this parser is a bit different. The order of the string doesn't matter and it parses the longest match, and not the first match. I prefer not to mix them.

I still have a bug to fix though, and I will add tests that cover it.

satabin:
The PR is now ready , I think. I ran the benchmarks and here are the results:

```
[info] Benchmark                         Mode  Cnt   Score   Error  Units
[info] StringInBenchmarks.oneOfParse     avgt   25  38,604 ± 3,312  ns/op
[info] StringInBenchmarks.stringInParse  avgt   25  22,861 ± 0,162  ns/op
```

bench/src/main/scala/cats/parse/bench/StringInBench.scala
```diff
@@ -0,0 +1,61 @@
...
+
+import cats.parse.Parser
+import java.util.concurrent.TimeUnit
+import org.openjdk.jmh.annotations._
+
+@State(Scope.Benchmark)
+@BenchmarkMode(Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+class StringInBenchmarks {
+  val text: String =
```
johnynek, line 32: instead can we have a list of all the strings that match, and also some that don't match, and benchmark `foreach(parseAll(_))...`

bench/src/main/scala/cats/parse/bench/StringInBench.scala
```diff
@@ -0,0 +1,61 @@
...
+        Parser.string1("foo"),
+        Parser.string1("bar")
+      )
+    )
+
+  @Benchmark
+  def stringInParse(): Unit =
+    stringIn.parseAll(text) match {
+      case Right(()) => ()
+      case Left(e) => sys.error(e.toString)
```
johnynek, line 51: I think we should benchmark failures as well, since real parsers do have failures.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -476,22 +478,33 @@ object Parser extends ParserInstances {
             (left, right) match {
               case (Str(_, s1), Str(_, s2)) => s1.compare(s2)
               case (Str(_, _), _) => -1
-              case (InRange(_, _, _), Str(_, _)) => 1
+              case (OneStr(_, _), Str(_, _)) => 1
+              case (OneStr(_, s1), OneStr(_, s2)) =>
+                val s = s1.lengthCompare(s2.size)
+                if (s == 0) s1.compare(s2)
```
johnynek, line 484: why not just `s1.compare(s2)` here?

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -100,6 +100,12 @@ object ParserGen {
       Gen.const(GenT(Parser.anyChar))
     )
 
+  val stringIn1: Gen[GenT[Parser1]] =
+    Arbitrary.arbitrary[List[String]].map { cs =>
+      if (cs.size < 2 || cs.exists(_.isEmpty)) GenT(Parser.fail: Parser1[Unit])
```
johnynek, line 105: I don't think you need the cs.size check anymore since you special cased the constructor method.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -562,6 +572,13 @@ class ParserTest extends munit.ScalaCheckSuite {
     parseTest(Parser.product(fooP, barP), "foobar", ((), ()))
   }
 
+  test("longest match stringIn1") {
+    parseTest(Parser.stringIn1(List("foo", "foobar", "foofoo", "foobat")).string, "foo", "foo")
+    parseTest(Parser.stringIn1(List("foo", "foobar", "foofoo", "foobat")).string, "foobat", "foobat")
+    parseTest(Parser.stringIn1(List("foo", "foobar", "foofoo", "foobat")).string, "foot", "foo")
+    parseTest(Parser.stringIn1(List("foo", "foobar", "foofoo", "foobat")).string, "foobal", "foo")
+  }
+
```
johnynek, line 585: can we do a law such as "if oneOf matches with all string1 parsers, then stringIn1 would match"

something like:
```scala
forAll { (ss0: List[String], toParse: String) =>
  val ss = ss0.filterNot(_.isEmpty)
  val oneOfs = P.oneOf1(ss.map(P.string1))
  val stringIn = P.stringIn1(ss)
  if (oneOfs.parse(toParse).isRight) assert(stringIn.parse(toParse).isRight)
}
```
or something like that.

I'd like to find more laws like that so we can stress it a bit harder.

Also a law that stringIn1(s :: Nil) == string1(s)
etc...

core/shared/src/main/scala/cats/parse/RadixNode.scala
```diff
@@ -0,0 +1,104 @@
...
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package cats.parse
+
+import cats.data.NonEmptyList
+
+import scala.annotation.tailrec
+
+class RadixNode(
```
johnynek, line 28: I see you are using an Array here to avoid the boxing of using a map. Another approach would be to use scala.collection.immutable.IntMap and lift the Char into an Int, and then IntMap won't box.

That will avoid the binary search, which might be a significant win.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1724,6 +1751,39 @@ object Parser extends ParserInstances {
       null.asInstanceOf[A]
     }
 
+    final def stringIn1[A](radix: RadixNode, all: List[String], state: State): Unit = {
+      val startOffset = state.offset
+      var offset = state.offset
+      var tree = radix
+      var cont = offset < state.str.length
```
johnynek, line 1758: might be worth a minor optimization to capture state.str.length in a local val to make it clear to the compiler/jit that does not change.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1724,6 +1751,39 @@ object Parser extends ParserInstances {
...
+          } else {
+            cont = false
+          }
+        } else {
+          cont = false
+        }
+      }
+      if (lastMatch < 0) {
+        state.error = Chain.one(Expectation.OneStr(startOffset, all))
+        state.offset = startOffset
```
johnynek, line 1781: by resetting to the original offset this parser *does* backtrack. Is that what you want? The documentation says it does not.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1724,6 +1751,39 @@ object Parser extends ParserInstances {
...
+            if(children.word) lastMatch = offset
+          } else {
+            cont = false
+          }
+        } else {
+          cont = false
+        }
+      }
+      if (lastMatch < 0) {
+        state.error = Chain.one(Expectation.OneStr(startOffset, all))
```
johnynek, line 1780: I think it would be better to generate the subtree that failed and the offset that we failed at. So, if we have foobar, foobaz, foofoo, and we try to parse fooquux, see would point to quux and say we expected bar, baz or foo. That would be a better error message I think.

What do you think?

johnynek (changes requested):
This looks nice.

Can you post the results of the benchmark as you run them and update them? This helps us record also for the future if they are logged in the message of the PR.

bench/src/main/scala/cats/parse/bench/StringInBench.scala
satabin: Sure

bench/src/main/scala/cats/parse/bench/StringInBench.scala
satabin: Will do that

core/shared/src/main/scala/cats/parse/Parser.scala
satabin: Oops some leftovers from a previous attempt. I will change this.

core/shared/src/test/scala/cats/parse/ParserTest.scala
satabin: Not this case, this makes sure that we don't have the empty string in alternatives as this is a `Parser1`, it cannot accept the empty string.

core/shared/src/main/scala/cats/parse/RadixNode.scala
satabin: Indeed, let's see if it improves. The `Map` method also has the overhead of creating an `Option`, I will compare to the current approach.

core/shared/src/main/scala/cats/parse/Parser.scala
satabin: I like to see the strings as atoms which either match completely or not at all, so I like to see the error at the beginning of the atom. But I am fine either way in the end.

core/shared/src/test/scala/cats/parse/ParserTest.scala
johnynek: but you also do a size check in stringIn1, so it is checking if size is < 2 (if 0 it returns fail, if 1 it returns stringIn).

Also, since these could in principle be two duplicate items, I think it is better to remove the size check here and just let stringIn1 do its work.

satabin: Oh sorry I misread, of course this test is now not used anymore. I will remove it.

core/shared/src/test/scala/cats/parse/ParserTest.scala
satabin: Both would match but parsed result might differ, do you think this is a strong enough law?

core/shared/src/main/scala/cats/parse/Parser.scala
satabin: Documentation was wrong (and contradicting itself), I removed the last the last false claim.

core/shared/src/test/scala/cats/parse/ParserTest.scala
johnynek: any law is better than no law, which is the current state. Let's add this, but also any others you can find.

another law might be the direct law:

if any of these would match with string1, then stringIn will match and return the longest match. Something like that?

core/shared/src/main/scala/cats/parse/Parser.scala
satabin: I thought about it again, and a typical use case for this operator is to match enum-like names. As a user, I think I am more interested in knowing where it fails at the beginning of the matching of the names than somewhere in the middle of one of the possibilities.

For instance if I want to parse the name of a weekday, I'd have:
```scala
stringIn1(List("monday", "tuesday", "wednesday", "thursday", "firday", "saturday", "sunday"))
```

Let's say I made a type and wrote `thrusday`, it would feel weird to me to know that my possiblities where `hursday` and `uesday`.
Of course this means that the failure offset is closer to the matching error, but semantically speaking, the list in `stringIn1` is what I, as a user, defined as my possibilities, and would expect to see in error messages.

What do you think?

bench/src/main/scala/cats/parse/bench/StringInBench.scala
```diff
@@ -0,0 +1,56 @@
...
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+class StringInBenchmarks {
+  val inputs =
+    List("foofoo", "bar", "foobat", "foot", "foobar")
+
+  val stringIn = Parser.stringIn1(List("foo", "bar", "foobar", "foofoo", "foobaz"))
+
+  val oneOf =
+    Parser.oneOf1(
+      List(
```
johnynek, line 39: nit: can we use the style of using `::` to construct the args:
```scala
oneOf1(
  Parser.string1("foobar") ::
  Parser.string1("foobaz") ::
  ...
```
This makes using the List more ergonomic since you are just replacing , with :: (and slightly more efficient since you don't need to use the varargs constructors.

satabin:
I took your comments and feedback into account and pushed the new version. I kept the array based radix tree implementation due to the results of the benchmarks:

```
Arrays
Benchmark                         Mode  Cnt    Score   Error  Units
StringInBenchmarks.stringInParse  avgt   25  163,929 ± 4,582  ns/op

IntMap.get
Benchmark                         Mode  Cnt    Score   Error  Units
StringInBenchmarks.stringInParse  avgt   25  198,055 ± 3,762  ns/op

IntMap.contains + IntMap.apply
Benchmark                         Mode  Cnt    Score    Error  Units
StringInBenchmarks.stringInParse  avgt   25  231,887 ± 10,620  ns/op

oneOf
Benchmark                      Mode  Cnt    Score    Error  Units
StringInBenchmarks.oneOfParse  avgt   25  296,758 ± 22,775  ns/op
```

The `IntMap` implementation creates more objects (especially the `Option` for the result), or requires two lookups (contains+apply), which results in some overhead over the tree implementation.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -501,6 +502,7 @@ object Parser extends ParserInstances {
 
   object Expectation {
     case class Str(offset: Int, str: String) extends Expectation
+    case class OneStr(offset: Int, strs: List[String]) extends Expectation
```
johnynek, line 505: I wonder if this should be a Set[String] or SortedSet[String] actually, since the order does not matter.

Also, I wonder if we really need Str and OneStr, maybe we just combine them into one error case. Having too many error results makes it harder to present simple messages to the users.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -842,6 +852,18 @@ object Parser extends ParserInstances {
     }
   }
 
+  /** Parse the longest matching string between alternatives.
+    * The order of the strings does not matter.
+    *
+    * If no string matches, this parser results in an epsilon failure.
+    */
+  def stringIn1(strings: List[String]): Parser1[Unit] =
```
johnynek, line 860: since order does not matter here maybe we should take an Iterable[String] to make that clearer. List (or any Seq) tends to imply order does matter.

core/shared/src/main/scala/cats/parse/Parser.scala
johnynek: I think that makes sense. I guess one thing I'm concerned about is this list getting very big and seeing a giant list on the typo. Like, in your example, I might say, I would expect Thursday or Tuesday, but I don't want to see  Sunday in the list, personally. What do you think of that?

core/shared/src/main/scala/cats/parse/Parser.scala
johnynek: We need to also update def doesBacktrack since this parser does backtrack, so there is no need to wrap it with a Backtrack node if a user wants backtracking.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1796,6 +1853,15 @@ object Parser extends ParserInstances {
       override def parseMut(state: State): A = oneOf(ary, state)
     }
 
+    case class StringIn1(all: List[String]) extends Parser1[Unit] {
```
johnynek, line 1856: maybe we should accept a SortedSet[String] here, which makes it structurally deterministic, and then we can skip having sorted, and just all.toList will be a sorted list we can pass to the RadixNode. What do you think?

core/shared/src/main/scala/cats/parse/RadixNode.scala
johnynek: did you have a chance to try the IntMap approach? I would think for larger sets it would be a big win, but I could be wrong.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -581,6 +592,17 @@ class ParserTest extends munit.ScalaCheckSuite {
     parseTest(Parser.product(fooP, barP), "foobar", ((), ()))
   }
 
+  test("longest match stringIn1") {
+    parseTest(Parser.stringIn1(List("foo", "foobar", "foofoo", "foobat")).string, "foo", "foo")
```
johnynek, line 596: can we use a val to not repeat the parser here? It looks like we repeat it 4 times.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1778,4 +1800,36 @@ class ParserTest extends munit.ScalaCheckSuite {
...
+  }
+
+  property("string1(s) matches  => stringIn1(ss) matches if s in ss") {
+    forAll { (s: String, ss0: List[String], toParse: String) =>
+      val ss = ss0.filterNot(_.isEmpty)
+      val ss1 = Random.shuffle(s :: ss)
+      if (s.nonEmpty && Parser.string1(s).parse(toParse).isRight)
+        assert(Parser.stringIn1(ss1).parse(toParse).isRight)
+    }
+  }
```
johnynek, line 1834: isn't it also true that:
```
P.string1(s0).orElse(P.string1(s1)).parse(arg).toOption ==
  P.stringIn1(s0 :: s1 :: Nil).parse(arg).toOption
```
That is: if we union to string parsers we get the same as a stringIn?

Also, doesn't that also hold for two stringIn parsers?
```
P.stringIn1(s0).orElse(P.stringIn1(s1)).parse(arg).toOption ==
  P.stringIn1(s0 ::: s1).parse(arg).toOption
```

johnynek (changes requested):
This is coming along very nicely

core/shared/src/test/scala/cats/parse/ParserTest.scala
johnynek: Okay, I don't think these laws are true in general. I think they are only true if the first item in the orElse has no prefixes in the second. I think we can check for that in the law:
```scala
val noPrefix = left.forall { s => !right.exists(_.startsWith(s)) }
```
something like that.

finally, here is one more law:
```scala
forAll { (ss0: List[String], toParse: String) =>
  val ss = ss0.filterNot(_.isEmpty)
  val left = P.stringIn1(ss).parse(toParse).toOption
  val right = ss.filter(toParse.startsWith(_)).sortBy { s => -s.length }
  assertEquals(left, right.headOption)
}
```

satabin: If `s0` is a prefix of `s1` then result is different when `arg` matches `s1`. The thing is that the radix tree approach really doesn't care about the order of the alternatives, whereas the `orElse` approach does and takes the first one that matches, eagerly.

johnynek: Yes, that's why I suggested the modification above. Maybe our comments crossed in flight.

satabin: Oh I didn't se that when replying, my bad. I will add these laws.

johnynek:
after you add the laws and merge main, I'm ready to merge this.

Thanks.

Note #116 is a pretty big change you need to be a bit careful with (Parser1 became Parser and Parser became Parser0)

satabin:
> after you add the laws and merge main, I'm ready to merge this.
> 
> Thanks.
> 
> Note #116 is a pretty big change you need to be a bit careful with (Parser1 became Parser and Parser became Parser0)

I didn't have much time this weekend, I'll try to finish this tomorrow, and rebase everything properly.

satabin:
I think I took all your comments into account. For the `IntMap` approach, see the benchmark results.

satabin:
These are the latest results of the benchmark with the current code

```
Benchmark                         Mode  Cnt    Score   Error  Units
StringInBenchmarks.oneOfParse     avgt   25  325,339 ± 7,546  ns/op
StringInBenchmarks.stringInParse  avgt   25  154,768 ± 3,157  ns/op
```

johnynek (approved):
This looks really great!

Thanks for addressing all my comments. I really appreciate it.

This will really improve the library.

I plan to release in a day or two version 0.3.0 with this change.

PR #115, opened 2020-12-14
Add either Parser

regadas:
Can this be useful? not sure about the naming though.


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -112,6 +112,21 @@ sealed abstract class Parser[+A] {
...
+    * If this parser fails with an arresting error, the next parser
+    * won't be tried.
+    *
+    * Backtracking may be used on the left parser to allow the right
+    * one to pick up after any error, resetting any state that was
+    * modified by the left parser.
+    *
+    * This method is similar to Parser#orElse but returns Either.
+    */
+  def or[B](pb: Parser[B]): Parser[Either[A, B]] =
```
johnynek, line 127: one challenge is the convention that the Right is "right". So, I might say, returning `Parser[Either[B, A]]` instead.

That said, I know that that is also the case that lexically the left side has type A and the right side has type B, so it seems that Left and Right should match...

I'm somewhat inclined to make a new ADT that has First and Second, which side steps this issue, but then that requires users to use those.

Finally, I'm a bit concerned with the name since it is so close to orElse. I know the method is similar to orElse, so maybe that's not a problem.

what about eitherOr? That is an expression people use in English. It is longer, but it does make it clearer which one is adding Either.

johnynek (commented):
I really like this method, but I did make a couple of comments and I look forward to your reply.

core/shared/src/main/scala/cats/parse/Parser.scala
rossabaker: Does it have to be an infix method?  `P.either(pa, pb)` feels natural to me.  There'd need to be an `either0` or an `either1`, depending on the fate of #116.

I think we this inlined a few places in http4s already. I'm in favor of the general concept.

johnynek: I actually do like `P.either` actually... but it does not solve the question of Left result being the default, when Right is usually the canonical value.

PS: cats should just add a Result type with Err and Ok and copy rust. This Either business is a mistake when combined with one side usually/often having a bias....

johnynek: (and yes, I remember Xor, and I remember the debates about making Either right biased... I think Either should be right biased, but Result a-la rust is a better design than Xor).

johnynek: actually, isn't this maybe a PR for cats in Alternative of MonoidK + Functor (there isn't a type like that is there?)

regadas: Hi! sorry for the late reply! A lot of good comments here that left me with no really good answer actually.

Initially, I had `P.either` but then I thought that this would probably be used more as infix, similar to `orElse`, and kept it lexical for Left and Right. @johnynek very good point about the `right` bias and I think it should be `Parser[Either[B, A]]`. I'm lacking an alternative name to `or` that conveys better "error" meaning to the Left. Open for suggestions...

> cats should just add a Result type with Err and Ok and copy rust. This Either business is a mistake when combined with one side usually/often having a bias....

+1

> actually, isn't this maybe a PR for cats in Alternative or MonoidK + Functor (there isn't a type like that is there?)

not sure... maybe? but isn't `orElse` more in line with `Alternative`?

regadas: would `recover` be completely off?

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -862,6 +862,32 @@ object Parser extends ParserInstances {
...
+    * try the left parser instead.
+    *
+    * If the right parser fails with an arresting error, the left parser
+    * won't be tried.
+    *
+    * Backtracking may be used on the left parser to allow the right
+    * one to pick up after any error, resetting any state that was
+    * modified by the left parser.
+    */
+  def either[A, B](left: Parser[A], right: Parser[B]): Parser[Either[A, B]] =
```
regadas, line 875: There was interest in having these. Here I gave preference to the right parser as it's seen as the "success". However, this gave me some doubts regarding the parameter order.

Also, should we just prefer these over the infix ones or keep both?

I believe the infix ones are more explicit in regards to which parser is attempted first and thus the "right" side.

wdyt?

johnynek: 1. There are two ideas in tension here: 1 left to right reading order suggests left comes first, 2 convention of Either is that biases to Right.
2. What name should we have, if any, for a method.

In the method case, the "this" has to be the first one read. I tend to think the same for def either. So I guess I would break the normal Right biased-ness. I don't know.

For the method name, I'm pretty concerned about confusion with or and orElse. I still like my suggestion of eitherOr as a method.

regadas: Let's go with `eitherOr`, revert back to not be right biased and see how that looks.

regadas: I have to say that `eitherOr` grew on me.

regadas: @johnynek Add the 2 commits one with and without right bias. I think the latter looks good now.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -842,6 +862,32 @@ object Parser extends ParserInstances {
     }
   }
 
+  /** If this right parser fails to parse its input with an epsilon error,
+    * try the left parser instead.
+    *
+    * If the right parser fails with an arresting error, the left parser
+    * won't be tried.
+    *
+    * Backtracking may be used on the left parser to allow the right
```
johnynek, line 871: The backtracking is on the "right" isn't it?

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -842,6 +862,32 @@ object Parser extends ParserInstances {
...
+  def eitherOr[A, B](right: Parser[B], left: Parser[A]): Parser[Either[A, B]] =
+    oneOf(right.map(Right(_)) :: left.map(Left(_)) :: Nil)
+
+  /** If this right parser fails to parse its input with an epsilon error,
+    * try the left parser instead.
+    *
+    * If the right parser fails with an arresting error, the left parser
+    * won't be tried.
+    *
+    * Backtracking may be used on the left parser to allow the right
```
johnynek, line 884: The "right" one is the one may use backtracking.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -842,6 +862,32 @@ object Parser extends ParserInstances {
...
+    * one to pick up after any error, resetting any state that was
+    * modified by the left parser.
+    */
+  def eitherOr1[A, B](right: Parser1[B], left: Parser1[A]): Parser1[Either[A, B]] =
```
johnynek, line 888: What about saying first and second to avoid the confusion with names? The types will help us not get confused with the order of the args?

johnynek:
can you merge main?

regadas:
@johnynek merged. Thanks for the review!

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1782,4 +1782,16 @@ class ParserTest extends munit.ScalaCheckSuite {
       assertEquals(left.parse(str), right.parse(str))
     }
   }
+
+  property("Or parser works as expected") {
```
johnynek, line 1786: Can we also test the method version on Parser0?

johnynek (changes requested):

core/shared/src/test/scala/cats/parse/ParserTest.scala
regadas: Good point!

johnynek (approved):

johnynek:
Thanks. This looks great.

Issue #113, opened 2020-12-12
1's everywhere are cumbersome


martijnhoekstra:
This issue may be a bit premature/pretentious to file as an issue, but I figured it would be good to get it out early. The safety that `Parser1` and its combinators are great. So great, that while trying out the library, I find I use them pretty much all the time. Using a `Parser` or a method that creates a `Parser` is an exception.

Has it been considered doing the other way around, renaming `Parser` to `Parser0`, and `Parser1` to `Parser`, along with all `1` methods on object `Parser`? That would make the potentially non-consuming Parsers the exception not only in practice but also in naming.

johnynek:
This is a very interesting suggestion. Honestly, the Parser1 idea came after the library had somewhat been implemented, and I didn't reconsider your idea.

I agree that generally I use Parser1, basically only Parser (Parser0) for whitespace or parsing the inside of repeated structures (the result of rep).

Would you be willing to make a PR to explore this change?

martijnhoekstra:
Sure, I'll see if I can get that done this weekend.

johnynek:
@rossabaker you might also have an opinion here.

My bias in general is to live with warts and don't break users. That said, we may have very few users at this point and they may appreciate the change anyway.

There are also two questions:

1. Do we just rename the types
2. Do we also rename the methods so that we remove trailing 1 and instead use trailing 0 for parsers that can be empty.

johnynek:
@regadas id love for you to weigh in too. You've been quite active here.

rossabaker:
As one of the few users, it's version 0.2, the project was announced like a month ago, and I'd rather change a few things now and get it right than live with it wrong forever.  And Somebody™ could always write a scalafix.

> Do we also rename the methods so that we remove trailing 1 and instead use trailing 0 for parsers that can be empty.

I think it would be confusing if we didn't.

martijnhoekstra:
Sloppy WIP at https://github.com/typelevel/cats-parse/compare/main...martijnhoekstra:Parser0?expand=1

Preliminary looks of things

* Combinators look much nicer without the 1 everywhere
* Do the right thing by default
* Constructors on object `Parser0` look bad. They should probably be moved to a to be newly introduced `object Parser`. Note that you can't straight up rename it, because Impl is private and parseMut is protected

johnynek:
[...]

PR #116, opened 2020-12-14
Make 1 the default: rename Foo1 things to Foo, and the associated Foo to Foo0

martijnhoekstra:
To illustrate https://github.com/typelevel/cats-parse/issues/113

Moved the constructors and their helper wrappers to `object Parser` (in the easiest way possible, by loosening access), in the second to last commit, and taking things to their logical extreme in I removed `parse` and `parseAll` from `Parser0`: it makes little sense to run an end-parser on a string that may not consume any characters, 

That the intuition works out is shown in that none of the "real world" tests needed changing other than catching that some `Parser(1)`'s could be `Parser0`'s.

Todo: Fixup the fallout of submitting too early while by while writing up the body of the PR and still running tests.

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -1181,8 +1181,8 @@ object Parser {
       case s if Impl.alwaysSucceeds(s) => unit
       case _ =>
         Impl.unmap0(pa) match {
-          case Impl.StartParser0 => Impl.StartParser0
-          case Impl.EndParser0 => Impl.EndParser0
+          case Impl.StartParser => Impl.StartParser
+          case Impl.EndParser => Impl.EndParser
           case n @ Impl.Not(_) => n
           case p @ Impl.Peek(_) => p
           case other => Impl.Void0(other)
@@ -1277,11 +1277,11 @@ object Parser {
 
   /** succeeds when we are at the start
     */
-  def start: Parser0[Unit] = Impl.StartParser0
+  def start: Parser0[Unit] = Impl.StartParser
 
   /** succeeds when we are at the end
     */
-  def end: Parser0[Unit] = Impl.EndParser0
+  def end: Parser0[Unit] = Impl.EndParser
 
   /** If we fail, rewind the offset back so that
     * we can try other branches. This tends
@@ -1425,9 +1425,7 @@ object Parser {
     final def doesBacktrack(p: Parser0[Any]): Boolean =
       p match {
         case Backtrack0(_) | Backtrack(_) | AnyChar | CharIn(_, _, _) | Str(_) | IgnoreCase(_) |
-            Length(_) | StartParser0 | EndParser0 | Index | Pure(_) | Fail() | FailWith(_) | Not(
-              _
-            ) =>
+            Length(_) | StartParser | EndParser | Index | Pure(_) | Fail() | FailWith(_) | Not(_) =>
           true
         case Map(p, _) => doesBacktrack(p)
         case Map1(p, _) => doesBacktrack(p)
@@ -1517,7 +1515,7 @@ object Parser {
         case Defer0(fn) =>
           Defer0(() => unmap0(compute0(fn)))
         case Rep0(p, _) => Rep0(unmap(p), Accumulator0.unitAccumulator0)
-        case StartParser0 | EndParser0 | TailRecM0(_, _) | FlatMap0(_, _) =>
+        case StartParser | EndParser | TailRecM0(_, _) | FlatMap0(_, _) =>
           // we can't transform this significantly
           pa
       }
@@ -1671,7 +1669,7 @@ object Parser {
         Impl.string0(parser, state)
     }
 
-    case object StartParser0 extends Parser0[Unit] {
+    case object StartParser extends Parser0[Unit] {
       override def parseMut(state: State): Unit = {
         if (state.offset != 0) {
           state.error = Chain.one(Expectation.StartOfString(state.offset))
@@ -1680,7 +1678,7 @@ object Parser {
       }
     }
 
-    case object EndParser0 extends Parser0[Unit] {
+    case object EndParser extends Parser0[Unit] {
       override def parseMut(state: State): Unit = {
         if (state.offset != state.str.length) {
           state.error = Chain.one(Expectation.EndOfString(state.offset, state.str.length))
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -1214,12 +1214,12 @@ object Parser {
     */
   def string0(pa: Parser0[Any]): Parser0[String] =
     pa match {
-      case str @ Impl.StringP(_) => str
+      case str @ Impl.StringP0(_) => str
       case s1: Parser[_] => string(s1)
       case _ =>
         Impl.unmap0(pa) match {
           case Impl.Pure(_) | Impl.Index => emptyStringParser0
-          case notEmpty => Impl.StringP(notEmpty)
+          case notEmpty => Impl.StringP0(notEmpty)
         }
     }
 
@@ -1228,7 +1228,7 @@ object Parser {
     */
   def string(pa: Parser[Any]): Parser[String] =
     pa match {
-      case str @ Impl.StringP1(_) => str
+      case str @ Impl.StringP(_) => str
       case _ =>
         Impl.unmap(pa) match {
           case len @ Impl.Length(_) => len
@@ -1241,7 +1241,7 @@ object Parser {
             // these are really Parser[Nothing]
             // but scala can't see that, so we cast
             f.asInstanceOf[Parser[String]]
-          case notStr => Impl.StringP1(notStr)
+          case notStr => Impl.StringP(notStr)
         }
     }
 
@@ -1465,7 +1465,7 @@ object Parser {
           unmap0(p)
         case Select0(p, fn) =>
           Select0(p, unmap0(fn))
-        case StringP(s) =>
+        case StringP0(s) =>
           // StringP is added privately, and only after unmap0
           s
         case Void0(v) =>
@@ -1542,7 +1542,7 @@ object Parser {
           unmap(p)
         case Select(p, fn) =>
           Select(p, unmap0(fn))
-        case StringP1(s) =>
+        case StringP(s) =>
           // StringP is added privately, and only after unmap
           s
         case Void(v) =>
@@ -1659,12 +1659,12 @@ object Parser {
       str
     }
 
-    case class StringP[A](parser: Parser0[A]) extends Parser0[String] {
+    case class StringP0[A](parser: Parser0[A]) extends Parser0[String] {
       override def parseMut(state: State): String =
         Impl.string0(parser, state)
     }
 
-    case class StringP1[A](parser: Parser[A]) extends Parser[String] {
+    case class StringP[A](parser: Parser[A]) extends Parser[String] {
       override def parseMut(state: State): String =
         Impl.string0(parser, state)
     }
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -1055,17 +1055,17 @@ object Parser0 extends ParserInstances {
 
   /** Lazily create a Parser
     *  This is useful to create some recursive parsers
-    *  see Defer[Parser].fix
+    *  see Defer0[Parser].fix
     */
   def defer1[A](pa: => Parser[A]): Parser[A] =
-    Impl.Defer1(() => pa)
+    Impl.Defer(() => pa)
 
   /** Lazily create a Parser0
     *  This is useful to create some recursive parsers
-    *  see Defer[Parser].fix
+    *  see Defer0[Parser].fix
     */
   def defer[A](pa: => Parser0[A]): Parser0[A] =
-    Impl.Defer(() => pa)
+    Impl.Defer0(() => pa)
 
   /** A parser that always fails with an epsilon failure
     */
@@ -1511,8 +1511,8 @@ object Parser0 extends ParserInstances {
               if (u2 eq Parser0.unit) u1
               else SoftProd(u1, u2)
           }
-        case Defer(fn) =>
-          Defer(() => unmap(compute0(fn)))
+        case Defer0(fn) =>
+          Defer0(() => unmap(compute0(fn)))
         case Rep(p, _) => Rep(unmap1(p), Accumulator0.unitAccumulator0)
         case StartParser0 | EndParser0 | TailRecM(_, _) | FlatMap(_, _) =>
           // we can't transform this significantly
@@ -1593,8 +1593,8 @@ object Parser0 extends ParserInstances {
               if (u2 eq Parser0.unit) expect1(u1)
               else SoftProd1(u1, u2)
           }
-        case Defer1(fn) =>
-          Defer1(() => unmap1(compute(fn)))
+        case Defer(fn) =>
+          Defer(() => unmap1(compute(fn)))
         case Rep1(p, m, _) => Rep1(unmap1(p), m, Accumulator0.unitAccumulator0)
         case AnyChar | CharIn(_, _, _) | Str(_) | IgnoreCase(_) | Fail() | FailWith(_) | Length(_) |
             TailRecM1(_, _) | FlatMap1(_, _) =>
@@ -1954,18 +1954,18 @@ object Parser0 extends ParserInstances {
     @annotation.tailrec
     final def compute0[A](fn: () => Parser0[A]): Parser0[A] =
       fn() match {
-        case Defer1(f) => compute(f)
-        case Defer(f) => compute0(f)
-        case notDefer => notDefer
+        case Defer(f) => compute(f)
+        case Defer0(f) => compute0(f)
+        case notDefer0 => notDefer0
       }
     @annotation.tailrec
     final def compute[A](fn: () => Parser[A]): Parser[A] =
       fn() match {
-        case Defer1(f) => compute(f)
-        case notDefer => notDefer
+        case Defer(f) => compute(f)
+        case notDefer0 => notDefer0
       }
 
-    case class Defer1[A](fn: () => Parser[A]) extends Parser[A] {
+    case class Defer[A](fn: () => Parser[A]) extends Parser[A] {
       private[this] var computed: Parser0[A] = null
       override def parseMut(state: State): A = {
 
@@ -1982,7 +1982,7 @@ object Parser0 extends ParserInstances {
       }
     }
 
-    case class Defer[A](fn: () => Parser0[A]) extends Parser0[A] {
+    case class Defer0[A](fn: () => Parser0[A]) extends Parser0[A] {
       private[this] var computed: Parser0[A] = null
       override def parseMut(state: State): A = {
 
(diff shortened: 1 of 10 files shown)


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -296,107 +296,110 @@ sealed abstract class Parser[+A] {
...
-  def surroundedBy(b: Parser[Any]): Parser[A] =
+  def surroundedBy(b: Parser0[Any]): Parser0[A] =
     between(b, b)
 
   /** Internal (mutable) parsing method.
     *
     * This method should only be called internally by parser instances.
     */
-  protected def parseMut(state: Parser.Impl.State): A
+  private[parse] def parseMut(state: Parser.Impl.State): A
```
johnynek, line 321: why does this need to be `private[parse]`, this does weaken the encapsulation a bit.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -65,7 +65,7 @@ sealed abstract class Parser[+A] {
     *
     * To require the entire input to be consumed, see `parseAll`.
     */
-  final def parse(str: String): Either[Parser.Error, (String, A)] = {
+  final def parse0(str: String): Either[Parser.Error, (String, A)] = {
```
johnynek, line 68: can we not change the names of these methods?

core/shared/src/test/scala/cats/parse/Rfc5234Test.scala
```diff
@@ -107,15 +107,15 @@ class Rfc5234Test extends munit.ScalaCheckSuite {
       )
       .map(_.mkString)
     forAll(genLwsp) { (s: String) =>
-      assertEquals(Rfc5234.lwsp.parseAll(s), Right(()))
+      assertEquals(Rfc5234.lwsp.parseAll0(s), Right(()))
```
johnynek, line 110: can we just call parseAll here, I think by the time you are parsing, you don't really care about the distinction.

johnynek (commented):
A few comments...

I'm still digesting how big the change is. It seems like almost every line touching the API will change.

I'd love to get unanimity from all reviewers this is worth it (@rossabaker @mpilquist @regadas )

core/shared/src/main/scala/cats/parse/Parser.scala
martijnhoekstra: Yes, I still want to try harder to narrow access back to the minimum. This was the easy way out to have a compiling PR showing the external API.

core/shared/src/main/scala/cats/parse/Parser.scala
martijnhoekstra: This is the "taking it to the extreme" part I mean in the main comment. I'm not married to the idea. The thought was, there are very few grammars that accept the empty string. Not calling a combinator with them but actually using them to call parse on could have an "alerting" name on them. Coincidentally, it did lead to finding Parsers being widened to Parser1 where they didn't need to.

The ugly parse0/parseAll0 show up here only in testing the subgrammar for white space, which is exactly the sore thumb effect it had in mind: careful, this might consume no input.

But like I said, I'm not married to the idea, and happy to take that commit out.

rossabaker:
The MiMa failures in the build would go away with a `baseVersion := "0.3"`.

What signfiicant adopters are we aware of?  Bosatsu, http4s, anything else?  Tweet and gitter about it and invite howls of pain?

I agree taking backward compatibility seriously, but I think we take it seriously by finding an API we're happy to live with long into the future than by living with something suboptimal to avoid a scalafix/sed for a few early adopters.  Judging it from the example parsers in the repo, I see more suffixes removed than added, and it feels better me.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -835,91 +836,91 @@ object Parser extends ParserInstances {
...
+    Impl.mergeCharIn[A, Parser0[A]](flat) match {
       case Nil => fail
       case p :: Nil => p
-      case two => Impl.OneOf(two)
+      case two => Impl.OneOf0(two)
     }
   }
 
-  private[this] val emptyStringParser: Parser[String] =
+  private[this] val emptyStringParser0: Parser0[String] =
```
martijnhoekstra, line 846: todo: Undo

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -238,30 +238,30 @@ sealed abstract class Parser[+A] {
...
-    *     val c2: Parser1[Int] = p.with1 <* p1
+    *     val c1: Parser0[Int]  = p <* p1
+    *     val c2: Parser[Int] = p.with1 <* p1
     *
-    * Without using `with1`, these methods will return `Parser` values
-    * since they are not known to return `Parser1` values instead.
+    * Without using `with1`, these methods will return `Parser0` values
+    * since they are not known to return `Parser` values instead.
     */
   def with1: Parser.With1[A] =
```
johnynek, line 266: this is maybe the only 1 name left... I guess it is better than withNot0 or something, but still it feels like it sticks out. I guess we should keep it to minimize change.

core/shared/src/main/scala/cats/parse/Parser.scala
johnynek: If we change this back to `protected` it still works, right?

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -438,59 +438,60 @@ sealed abstract class Parser1[+A] extends Parser[A] {
     * an epsilon failure, the parsed values (if any) are returned in a
     * list as a successful parse.
     */
-  def rep: Parser[List[A]] =
-    Parser.rep(this)
+  def rep0: Parser0[List[A]] =
```
johnynek, line 441: if we are breaking people should we just unify rep0 and repAs here? Maybe in the companion we remove rep0 and just call repAs variants, but leave the methods for convenience?

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -438,59 +438,60 @@ sealed abstract class Parser1[+A] extends Parser[A] {
...
+  /** This method overrides `Parser0#soft` to refine the return type.
     */
-  override def soft: Parser.Soft10[A] =
-    new Parser.Soft10(this)
+  override def soft: Parser.Soft[A] =
+    new Parser.Soft(this)
 }
 
-object Parser extends ParserInstances {
+object Parser0 extends ParserInstances {}
```
johnynek, line 493: okay, I guess we need this because of the way implicit resolution works. Do we need the `{}`? Can we add a comment to explain to readers the design?

core/shared/src/main/scala/cats/parse/Parser.scala
johnynek: I think the new name is fine, btw.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -835,91 +836,91 @@ object Parser extends ParserInstances {
...
+          case _ => Impl.Prod0(first, second)
         }
     }
 
-  /** product with the first argument being a Parser1
+  /** product with the first argument being a Parser
     */
-  def product10[A, B](first: Parser1[A], second: Parser[B]): Parser1[(A, B)] =
-    Impl.Prod1(first, second)
+  def product[A, B](first: Parser[A], second: Parser0[B]): Parser[(A, B)] =
```
johnynek, line 917: can we just keep the name product10 here, I think methods like this with several variants it is fine to still use the 1 to make it clearer.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -928,79 +929,79 @@ object Parser extends ParserInstances {
...
+  /** softProduct with the first argument being a Parser
     *  A soft product backtracks if the first succeeds and the second
     *  is an epsilon-failure. By contrast product will be a failure in
     *  that case
     *
     *  see @Parser.soft
     */
-  def softProduct10[A, B](first: Parser1[A], second: Parser[B]): Parser1[(A, B)] =
-    Impl.SoftProd1(first, second)
+  def softProduct[A, B](first: Parser[A], second: Parser0[B]): Parser[(A, B)] =
```
johnynek, line 950: rather keep 10 here.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1040,62 +1041,62 @@ object Parser extends ParserInstances {
     *  flatMap always has to allocate a parser, and the
     *  parser is less amenable to optimization
     */
-  def tailRecM[A, B](init: A)(fn: A => Parser[Either[A, B]]): Parser[B] =
+  def tailRecM[A, B](init: A)(fn: A => Parser0[Either[A, B]]): Parser0[B] =
```
johnynek, line 1044: shouldn't this be def tailRecM0 to follow your other changes, and let tailRec be the 1 variant?

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1040,62 +1041,62 @@ object Parser extends ParserInstances {
...
-  /** tail recursive monadic flatMaps on Parser1
+  /** tail recursive monadic flatMaps on Parser
     * This is a rarely used function, but needed to implement cats.FlatMap
     *  Avoid this function if possible. If you can
     *  instead use product, ~, *>, or <* use that.
     *  flatMap always has to allocate a parser, and the
     *  parser is less amenable to optimization
     */
-  def tailRecM1[A, B](init: A)(fn: A => Parser1[Either[A, B]]): Parser1[B] =
+  def tailRecM1[A, B](init: A)(fn: A => Parser[Either[A, B]]): Parser[B] =
```
johnynek, line 1054: def tailRecM

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1407,41 +1408,43 @@ object Parser extends ParserInstances {
 
     }
 
-  private object Impl {
+  private[parse] object Impl {
```
johnynek, line 1411: can we revert this to just private?

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1658,17 +1661,17 @@ object Parser extends ParserInstances {
       str
     }
 
-    case class StringP[A](parser: Parser[A]) extends Parser[String] {
+    case class StringP[A](parser: Parser0[A]) extends Parser0[String] {
```
johnynek, line 1664: should this be StringP0 for consistency?

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1658,17 +1661,17 @@ object Parser extends ParserInstances {
...
       override def parseMut(state: State): String =
-        Impl.string(parser, state)
+        Impl.string0(parser, state)
     }
 
-    case class StringP1[A](parser: Parser1[A]) extends Parser1[String] {
+    case class StringP1[A](parser: Parser[A]) extends Parser[String] {
```
johnynek, line 1669: just StringP?

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1658,17 +1661,17 @@ object Parser extends ParserInstances {
...
       override def parseMut(state: State): String =
-        Impl.string(parser, state)
+        Impl.string0(parser, state)
     }
 
-    case object StartParser extends Parser[Unit] {
+    case object StartParser0 extends Parser0[Unit] {
```
johnynek, line 1674: I don't think the name needs to change here, there is only one StartParser, similar for EndParser

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1741,21 +1744,21 @@ object Parser extends ParserInstances {
...
-    case class FailWith[A](message: String) extends Parser1[A] {
+    case class FailWith[A](message: String) extends Parser[A] {
       override def parseMut(state: State): A = {
         state.error = Chain.one(Expectation.FailWith(state.offset, message));
         null.asInstanceOf[A]
       }
     }
 
-    final def oneOf[A](all: Array[Parser[A]], state: State): A = {
+    final def oneOf0[A](all: Array[Parser0[A]], state: State): A = {
```
johnynek, line 1761: I don't think this name needs to change, there is only one internal implementation.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1939,41 +1942,41 @@ object Parser extends ParserInstances {
...
 
-    case class TailRecM[A, B](init: A, fn: A => Parser[Either[A, B]]) extends Parser[B] {
+    case class TailRecM[A, B](init: A, fn: A => Parser0[Either[A, B]]) extends Parser0[B] {
       private[this] val p1 = fn(init)
 
       override def parseMut(state: State): B = Impl.tailRecM(p1, fn, state)
     }
 
-    case class TailRecM1[A, B](init: A, fn: A => Parser1[Either[A, B]]) extends Parser1[B] {
+    case class TailRecM1[A, B](init: A, fn: A => Parser[Either[A, B]]) extends Parser[B] {
```
johnynek, line 1951: I think this should be TailRecM and the the other should be TailRecM0

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -2233,67 +2236,67 @@ object Parser extends ParserInstances {
 }
 
 abstract class ParserInstances {
```
johnynek, line 2238: since there are two companions now, is there any reason not to just make this `object Parser0`?

Also, should we alias all the def foo0 methods in here so people can do P0.string or P.string rather than P.string0 which is a bit weird.

johnynek (commented):
I'm pretty convinced this is the right move. One thing here is that the safer option: parsers that consume at least 1 thing are the easier to write. That is as it should be. It shouldn't feel like a tax to write a better parser. The tax should come when using the weaker parser (something that might return nothing).

core/shared/src/main/scala/cats/parse/Parser.scala
martijnhoekstra: along with `flatMap01` (because there also is flatMap (which was flatMap10) and flatMap0 (which was flatMap), `tailRecM1` which I just didn't see right away and should be changed for consistency, even though it's barely used in user code.

core/shared/src/main/scala/cats/parse/Parser.scala
martijnhoekstra: a find-and-replace mishap

core/shared/src/main/scala/cats/parse/Parser.scala
martijnhoekstra: For the first, I'll try it out. 

For the second, I don't have a strong opinion either way. Having a `P0.foo` as an alias for `P.foo0` meaning something subtly but breakingly different from `P0.foo` is a bit of a foot gun. I'm not sure if the added convenience is worth it.

johnynek: I don't follow how they would be different. Can you explain? I am proposing they point to exactly the same code.

martijnhoekstra: I meant to say the difference between P.foo and P0.foo being the foot gun, but in style I mistyped it.

core/shared/src/main/scala/cats/parse/Parser.scala
martijnhoekstra: It's now accessed from object Parser, which is a companion of a subclass. Can you do that? IMO spec says yes, compiler says no. https://github.com/scala/bug/issues/2439 says working as intended.

core/shared/src/main/scala/cats/parse/Parser.scala
martijnhoekstra: No, it's a member of `Parser`, and is accessed in `Parser0` in `parse`, `parseAll`, `?` and `parseMut`.

johnynek:
Is this ready to merge you think?

martijnhoekstra:
flatMap and (soft)product haven't been renamed back to the 10 versions yet (unless you changed your mind about that) and I haven't looked into unifying rep and repAs yet. A `RepParser[A] extends Parser[Seq[A]]` with methods to set sep, accumulator, min and max also sounds nice, with the current methods as shortcuts creating them. But it's orthogonal to this change I think.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -174,8 +174,8 @@ sealed abstract class Parser[+A] {
     * one to pick up after any error, resetting any state that was
     * modified by the left parser.
     */
-  def orElse[A1 >: A](that: Parser[A1]): Parser[A1] =
-    Parser.oneOf(this :: that :: Nil)
+  def orElse0[A1 >: A](that: Parser0[A1]): Parser0[A1] =
```
regadas, line 177: I was just thinking... could we rename this to just `orElse`? wdyt?

martijnhoekstra: The result will be that `Parser` has this inherited `orElse`, and an overloaded `orElse` with `Parser` as an argument and returning a `Parser`. After making the change and search-and-replacing orElse0 with orElse, all the tests still pass.

So I think it's safe to say yes, we could.

I have no strong opinion either way on whether we should and whether we should in this particular PR.

The tension is between the convenience of just being able to say `.orElse` and getting the most specific result possible depending on the receiver and argument types and the obviousness of which method is called when one has the 0.

cc @johnynek

johnynek (approved):
This looks great. I'm inclined to merge now. And strong objections?

To me, the main arguments are:

1. Almost all parsers (basically except for whitespace) parse at least one character.
2. This makes doing the more constrained and safer thing easier and adds a tax to putting less constraint, which is the right bias IMO.

martijnhoekstra:
I have a WIP for a scalafix too: https://github.com/martijnhoekstra/catsparsescalafix

It's not up to date for the latest changes yet, doesn't replace some instance methods yet (for some reason I haven't figured out yet), and sometimes leaves unused imports and sometimes clashes imports, but it greatly helps in migrating: https://github.com/johnynek/bosatsu/compare/master...martijnhoekstra:fixedCompile?expand=1 was just a few manual edits from the scalafix.

johnynek:
Can you add an MD file on how to use it to migrate. It would be awesome if we could figure out how to tell @scala-steward to use that scala fix to upgrade people too.

fthomas:
> It would be awesome if we could figure out how to tell @scala-steward to use that scala fix to upgrade people too.

Documentation for that is here: https://github.com/scala-steward-org/scala-steward/blob/master/docs/scalafix-migrations.md#adding-migration-rules-to-scala-steward

johnynek:
@martijnhoekstra can you merge main back in and I won't add anything else until this PR lands.

Then it looks like we can send a PR to scalasteward to point to the scalafix you have which hopefully we can get good enough that it is a very minor edit needed for any users.

Thanks a ton for this. It was a great idea and a great cleanup.

martijnhoekstra:
@johnynek there is nothing to merge -- I rebased earlier today.

johnynek:
Without any objections I'll merge in the morning.

regadas (approved):
LGTM, no strong objections on my side. I would like to revisit the `orElse` subject though.

johnynek:
Thanks all! I think this is a great change to make the library easier to use and steer us more towards non-empty parsers.

PR #122, opened 2020-12-19
Fix string unmap with Parser1

regadas:
Just noticed that for `string` it unmap'ing 2 times. not sure if I'm missing anything.


johnynek:
Thanks!

PR #125, opened 2020-12-22
Merge adjacent StringIn nodes in oneOf

johnynek:
The main idea here is to merge adjacent Str and StringIn nodes in order to take advantage of the log vs linear complexity of matching that the RadixNode offers

cc @satabin

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -911,16 +911,27 @@ object Parser {
     * The order of the strings does not matter.
     *
     * If no string matches, this parser results in an epsilon failure.
+    *
+    * It is an error to pass the empty string here, if you
+    * need that see stringIn0
     */
-  def stringIn(strings: Iterable[String]): Parser[Unit] =
+  def stringIn(strings: Iterable[String]): Parser[String] =
     strings.toList.distinct match {
       case Nil => fail
-      case s :: Nil => string(s)
+      case s :: Nil => string(s).string
       case two =>
-        Impl.StringIn(
-          SortedSet(two: _*)
-        ) // sadly scala 2.12 doesn't have the `SortedSet.from` constructor function
+        Impl
+          .StringIn(
+            SortedSet(two: _*)
+          ) // sadly scala 2.12 doesn't have the `SortedSet.from` constructor function
+          .string
     }
+
+  /** Version of stringIn that allows the empty string
+    */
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -617,6 +617,18 @@ object Parser {
             Impl.rangesFor(ary).map { case (l, u) => InRange(o, l, u) }.toList
           }
           .toList
+      // merge the OneOfStr
+      val oosMerge: List[OneOfStr] =
```
johnynek, line 621: this is needed to pass the laws once we are merging Str and StringIn nodes.

core/shared/src/main/scala/cats/parse/RadixNode.scala
```diff
@@ -35,33 +36,34 @@ class RadixNode(
...
+    keys match {
+      case key :: keys =>
+        val prefixSize = commonPrefixLength(prefix, key)
+        if (prefixSize == 0) {
+          // no common prefix, group current suffixes together sorted again
+          groupByNonEmptyPrefix(
+            keys,
+            key,
+            NonEmptyList.one(key),
+            (prefix(0), prefix.tail, current.map(_.drop(prefix.size)).reverse) :: acc
```
johnynek, line 56: I changed this to have the tail of the prefix so we don't compare twice the head when matching.

core/shared/src/main/scala/cats/parse/RadixNode.scala
```diff
@@ -81,24 +86,21 @@ object RadixNode {
...
 
   private val leaf = new RadixNode(Array.empty, Array.empty, Array.empty, true)
 
-  private def commonPrefix(s1: String, s2: String): Int = {
-    @tailrec
-    def loop(idx: Int): Int =
-      if (idx >= s1.size || idx >= s2.size) {
-        idx
+  final def commonPrefixLength(s1: String, s2: String): Int = {
+    val len = Integer.min(s1.length, s2.length)
```
johnynek, line 93: minor optimization to do only one comparison below, and do it against a local not the result of a method call.

johnynek:
@regadas would you be able to review this one?

regadas:
@johnynek yea sorry took a few days off 😄

regadas (approved):
LGTM

johnynek:
Thanks for the review @regadas I have incorporated your suggestions.

PR #130, opened 2020-12-30
Add | operator

johnynek:
Adds a synonym for orElse, which is common in parsing libraries.

### DIFF
--- a/core/shared/src/test/scala/cats/parse/ParserTest.scala
+++ b/core/shared/src/test/scala/cats/parse/ParserTest.scala
@@ -1201,6 +1201,14 @@ class ParserTest extends munit.ScalaCheckSuite {
     forAll(ParserGen.gen0, ParserGen.gen0) { (p1, p2) =>
       assertEquals(p1.fa.orElse(p2.fa), p1.fa | p2.fa)
     }
+
+    forAll(ParserGen.gen, ParserGen.gen0) { (p1, p2) =>
+      assertEquals(p1.fa.orElse(p2.fa), p1.fa | p2.fa)
+    }
+
+    forAll(ParserGen.gen0, ParserGen.gen) { (p1, p2) =>
+      assertEquals(p1.fa.orElse(p2.fa), p1.fa | p2.fa)
+    }
   }
 
   test("charWhere(_ => true) == anyChar") {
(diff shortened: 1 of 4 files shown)


johnynek:
@martijnhoekstra @regadas 

Hey folks, take a look at this. My anxiety about having two orElse methods seemed to have some truth to it... It looks like on 2.12, you can't mix a Parser0 and a Parser with orElse (or `|`). It seems that it works if both arguments are the same type.

So, to me, that suggests we should add orElse0 for Parser0 (removing orElse on Parser0), and then make | the operator only work when both arguments are Parser.

What do you think?

martijnhoekstra:
That's really sad. Is it a bug in the overload resolution of 2.12 for abstract types, or does it happen with all types? I'll poke at it a bit this afternoon.

regadas:
Yea, from a quick look this seems to be an issue with the dependent type in `GenT`. 2.12 seems to have issues with `p2.A >: p1.A`

```scala
    forAll(ParserGen.gen, ParserGen.gen0) { (p1, p2) =>
      val asd = p1.fa.orElse(p2.fa.asInstanceOf[Parser0[p1.A]])
      assertEquals(asd, p1.fa | p2.fa.asInstanceOf[Parser0[p1.A]])
    }
```

regular stuff works fine
```scala
Parser.string("a") | Parser.string0("b")
```

regadas:
Given that this is 2.12 specific and "normal" use cases are fine, I'm wondering if this issue is worth the extra API juggling and `|` available only on `Parser`.

johnynek:
good catch @regadas -- this is really about those generators not the Parser0/Parser types. If I ascribe a type, it seems to work.

So, I think this should probably be okay to merge.

johnynek:
okay, this seems to pass and i guess was just a bug with the path dependent type.

Any objection to merge?

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1193,6 +1193,29 @@ class ParserTest extends munit.ScalaCheckSuite {
...
+      assertEquals(pa.orElse(pb), pa | pb)
+    }
+    forAll(ParserGen.gen0, ParserGen.gen0) { (p1, p2) =>
+      val pa: Parser0[Any] = p1.fa
+      val pb: Parser0[Any] = p2.fa
+      assertEquals(pa.orElse(pb), pa | pb)
+    }
+    forAll(ParserGen.gen, ParserGen.gen0) { (p1, p2) =>
+      val pa: Parser[Any] = p1.fa
+      val pb: Parser0[Any] = p2.fa
```
regadas, line 1209: yea the ascription to Any makes sense.

regadas (approved):
👍

PR #132, opened 2021-01-02
allow for max in rep

### LINKED ISSUE #97: rep with min and max
Related to #52, question was raised about parsers that repeat a min and a max number of times.  Here's a real-world case from RFC7321, which has a precision to the thousandths:

```
     qvalue = ( "0" [ "." 0*3DIGIT ] )
            / ( "1" [ "." 0*3("0") ] )
```

It's not real common, but they're out there.

martijnhoekstra:
Add max parameter for rep(0) and repAs(0).

Fixes #97

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -2324,15 +2324,19 @@ object Parser {
       private[this] val ignore: B = null.asInstanceOf[B]
 
       override def parseMut(state: State): B = {
+        // first parse one, so we can initialize the appender with that value
+        // then do the rest, with min -> min - 1 and
+        // maxMinusOne -> maxMinusOne - 1 or Int.MaxValue as "forever" sentinel
         val head = p1.parseMut(state)
-        def maxRemaining = if (maxMinusOne == Int.MaxValue) Int.MaxValue else maxMinusOne - 1
+        def maxRemainingMinusOne =
+          if (maxMinusOne == Int.MaxValue) Int.MaxValue else maxMinusOne - 1
         if (state.error ne null) ignore
         else if (state.capture) {
           val app = acc1.newAppender(head)
-          if (repCapture(p1, min - 1, maxRemaining, state, app)) app.finish()
+          if (repCapture(p1, min - 1, maxRemainingMinusOne, state, app)) app.finish()
           else ignore
         } else {
-          repNoCapture(p1, min - 1, maxRemaining, state)
+          repNoCapture(p1, min - 1, maxRemainingMinusOne, state)
           ignore
         }
       }
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -520,10 +520,8 @@ sealed abstract class Parser[+A] extends Parser0[A] {
   def rep(min: Int): Parser[NonEmptyList[A]] =
     Parser.repAs(this, min = min)
 
-  def rep(min: Int, max: Int): Parser[NonEmptyList[A]] = {
-    require(min >= 1, s"min should be >= 1 but was $min")
+  def rep(min: Int, max: Int): Parser[NonEmptyList[A]] =
     Parser.repAs(this, min = min, max = max)
-  }
 
   /** This method overrides `Parser0#between` to refine the return type
     */
@@ -811,6 +809,9 @@ object Parser {
 
     def repAs[B](min: Int, max: Int)(implicit acc: Accumulator[A, B]): Parser[B] =
       Parser.repAs(self, min = min, max = max)(acc)
+
+    def repExactlyAs[B](times: Int)(implicit acc: Accumulator[A, B]): Parser[B] =
+      Parser.repAs(self, min = times, max = times)(acc)
   }
 
   /** Don't advance in the parsed string, just return a
@@ -1002,26 +1003,55 @@ object Parser {
     Impl.Length(len)
 
   /** Repeat this parser 0 or more times
-    * note: this can wind up parsing nothing
+    *
+    * @note this can wind up parsing nothing
     */
   def repAs0[A, B](p1: Parser[A])(implicit acc: Accumulator0[A, B]): Parser0[B] =
     Impl.Rep0(p1, Int.MaxValue, acc)
 
-  /** Repeat this parser 0 or more times
-    * note: this can wind up parsing nothing
+  /** Repeat this parser 0 or more times, but no more than `max`
+    *
+    * @throws IllegalArgumentException if max < 1
+    *
+    * @note this can wind up parsing nothing
     */
-  def repAs0[A, B](p1: Parser[A], max: Int)(implicit acc: Accumulator0[A, B]): Parser0[B] =
+  def repAs0[A, B](p1: Parser[A], max: Int)(implicit acc: Accumulator0[A, B]): Parser0[B] = {
+    require(max >= 1, s"max should be >= 1, was $max")
     Impl.Rep0(p1, max - 1, acc)
+  }
 
-  /** Repeat this parser 1 or more times
+  /** Repeat this parser `min` or more times
+    *
+    * The parser fails if it can't match at least `min` times
+    *
+    * @throws IllegalArgumentException if min < 1
     */
-  def repAs[A, B](p1: Parser[A], min: Int)(implicit acc: Accumulator[A, B]): Parser[B] =
+  def repAs[A, B](p1: Parser[A], min: Int)(implicit acc: Accumulator[A, B]): Parser[B] = {
+    require(min >= 1, s"min should be >= 1, was $min")
     Impl.Rep(p1, min, Int.MaxValue, acc)
+  }
 
-  /** Repeat this parser 1 or more times
+  /** Repeat this parser `min` or more times, but no more than `max`
+    *
+    * The parser fails if it can't match at least `min` times
+    * After repeating the parser `max` times, the parser completes succesfully
+    *
+    * @throws IllegalArgumentException if min < 1 or max < min
     */
-  def repAs[A, B](p1: Parser[A], min: Int, max: Int)(implicit acc: Accumulator[A, B]): Parser[B] =
+  def repAs[A, B](p1: Parser[A], min: Int, max: Int)(implicit acc: Accumulator[A, B]): Parser[B] = {
+    require(min >= 1, s"min should be >= 1, was $min")
+    require(max >= min, s"max should be >= min, but $max < $min")
     Impl.Rep(p1, min, max - 1, acc)
+  }
+
+  /** Repeat the parser exactly `times` times
+    *
+    * @throws IllegalArgumentException if times < 1
+    */
+  def repExactlyAs[A, B](p: Parser[A], times: Int)(implicit acc: Accumulator[A, B]): Parser0[B] = {
+    require(times >= 1, s"times should be >= 1, was $times")
+    Impl.Rep(p, times, times - 1, acc)
+  }
 
   /** Repeat 1 or more times with a separator
     */
@@ -2248,7 +2278,6 @@ object Parser {
         state: State,
         append: Appender[A, B]
     ): Boolean = {
-      //require(max >= 0)
       var offset = state.offset
       var cnt = 0
       //maxMinusOne == Int.MaxValue is a sentinel value meaning "forever"
@@ -2549,6 +2578,13 @@ object Parser0 {
 
       def mapFilter[A, B](fa: Parser0[A])(f: A => Option[B]): Parser0[B] =
         fa.mapFilter(f)
+
+      override def replicateA[A](n: Int, fa: Parser0[A]): Parser0[List[A]] = {
+        fa match {
+          case p: Parser[A] if n >= 1 => Parser.repExactlyAs(p, n)
+          case _ => super.replicateA(n, fa)
+        }
+      }
 
       override def filter[A](fa: Parser0[A])(fn: A => Boolean): Parser0[A] =
         fa.filter(fn)
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -2299,7 +2299,8 @@ object Parser {
       }
     }
 
-    case class Rep0[A, B](p1: Parser[A], maxMinusOne: Int, acc: Accumulator0[A, B]) extends Parser0[B] {
+    case class Rep0[A, B](p1: Parser[A], maxMinusOne: Int, acc: Accumulator0[A, B])
+        extends Parser0[B] {
       private[this] val ignore: B = null.asInstanceOf[B]
 
       override def parseMut(state: State): B = {
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -500,7 +500,7 @@ sealed abstract class Parser[+A] extends Parser0[A] {
     else this.repAs(min)
 
   def rep0(min: Int, max: Int): Parser0[List[A]] =
-    if (min == 0) rep0
+    if (min == 0) this.repAs0(max)
     else this.repAs(min, max)
 
   /** Use this parser to parse one-or-more values.
@@ -799,6 +799,9 @@ object Parser {
   implicit final class ParserMethods[A](private val self: Parser[A]) extends AnyVal {
     def repAs0[B](implicit acc: Accumulator0[A, B]): Parser0[B] =
       Parser.repAs0(self)(acc)
+
+    def repAs0[B](max: Int)(implicit acc: Accumulator0[A, B]): Parser0[B] =
+      Parser.repAs0(self, max)(acc)
 
     def repAs[B](implicit acc: Accumulator[A, B]): Parser[B] =
       Parser.repAs(self, min = 1)(acc)
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -1011,17 +1011,17 @@ object Parser {
     * note: this can wind up parsing nothing
     */
   def repAs0[A, B](p1: Parser[A], max: Int)(implicit acc: Accumulator0[A, B]): Parser0[B] =
-    Impl.Rep0(p1, max, acc)
+    Impl.Rep0(p1, max - 1, acc)
 
   /** Repeat this parser 1 or more times
     */
   def repAs[A, B](p1: Parser[A], min: Int)(implicit acc: Accumulator[A, B]): Parser[B] =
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -499,6 +499,10 @@ sealed abstract class Parser[+A] extends Parser0[A] {
     if (min == 0) rep0
     else this.repAs(min)
 
+  def rep0(min: Int, max: Int): Parser0[List[A]] =
+    if (min == 0) rep0
```
johnynek, line 503: Isn't this discarding max? Can we add a test that would have caught this?

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -2213,16 +2236,21 @@ object Parser {
...
         state: State,
         append: Appender[A, B]
     ): Boolean = {
+      //require(max >= 0)
       var offset = state.offset
       var cnt = 0
-
-      while (true) {
+      //max == Int.MaxValue is a sentinel value meaning "forever"
+      val maxMinusOne = if (max == Int.MaxValue) max else max - 1
```
johnynek, line 2252: Can we do this higher up? As it is, if you really mean Int.MaxValue you get infinity. If you do it higher up you would pass Int.MaxValue - 1 for that case and Int.MaxValue will be reserved for infinity.

That all said, maybe we should use Longs internally since 2 Billion really isn't totally enormous. Long.MaxValue is like infinity because no computer we are ever going to use will be able to run through that counter in any human scale time.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -2271,36 +2299,39 @@ object Parser {
...
+    case class Rep[A, B](p1: Parser[A], min: Int, max: Int, acc1: Accumulator[A, B])
+        extends Parser[B] {
       if (min < 1) throw new IllegalArgumentException(s"expected min >= 1, found: $min")
 
       private[this] val ignore: B = null.asInstanceOf[B]
 
       override def parseMut(state: State): B = {
         val head = p1.parseMut(state)
-
+        def maxRemaining = if (max == Int.MaxValue) max else max - 1
```
johnynek, line 2327: Feels like we are doubling this up. I'm confused how this is here and in the rep methods above. Can we do it in only one place?

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1086,6 +1108,66 @@ class ParserTest extends munit.ScalaCheckSuite {
...
+      {
+        val (min, max) = minmax
+        genP.fa.rep0(min, max).parse(str).foreach { case (_, l) =>
+          assert(l.length <= max)
+        }
+      }
+    }
+  }
+
+  object Counter extends Accumulator0[Unit, Int] {
```
johnynek, line 1141: This is actually an interesting instance (maybe with Long) to make implicit.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1086,6 +1108,66 @@ class ParserTest extends munit.ScalaCheckSuite {
...
+  }
+
+  property("repAs parses max entries when more is available") {
+    forAll(Gen.choose(1, 100)) { (n: Int) =>
+      val toBeConsumed = "a" * n * 2
+      val p = Parser.char('a').repAs(min = n, max = n)(Counter)
+      val parsed = p.parse(toBeConsumed)
+      assertEquals(parsed, Right(("a" * n, n)))
+    }
+  }
```
johnynek, line 1169: Can we add a test that we parse exactly N when min = N and max = N?

johnynek:
Also: maybe we should have an exactly method parsing exactly a given number (min == max). I think this can also override replicateA in Applicative.

core/shared/src/test/scala/cats/parse/ParserTest.scala
martijnhoekstra: In test, or globally?

rep parsers can't successfully rep more than Int.MaxValue times on any input String, since String is limited to Int.MaxValue characters. Is making it Long to future-proof it for maybe parsing against something like fs2.Stream in the future worth it?

core/shared/src/test/scala/cats/parse/ParserTest.scala
martijnhoekstra: Do the "rep0/rep/repAs0/repAs parses max entries when available" tests cover this?

core/shared/src/main/scala/cats/parse/Parser.scala
martijnhoekstra: Yes, it was. The property test that tests p.rep(min, max) results in no more than max results didn't catch this, even locally after running it a bunch of times in a row. There is an explicit test for this now.

Since the property test didn't catch it, I'm concerned it just eats CPU cycles and feedback loop latency.

core/shared/src/test/scala/cats/parse/ParserTest.scala
johnynek: I was thinking to put it in the companion to Accumulator0. Good point about Int vs Long.

core/shared/src/test/scala/cats/parse/ParserTest.scala
johnynek: I think in principle, yes, but as you say, we might not hit it very often (remember, with the scala check, most parsers will just fail to parse anyway, add repetition in there and maybe they almost never parse, so I think the tests may be much weaker than they appear at first glance).

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -2213,16 +2239,21 @@ object Parser {
...
+    /** capture parser p repeatedly, at least min times, at most max times
+      */
     final def repCapture[A, B](
         p: Parser[A],
         min: Int,
+        max: Int,
         state: State,
         append: Appender[A, B]
     ): Boolean = {
+      //require(max >= 0)
```
johnynek, line 2251: Is it valid for `max < min`? If so, what do we return? It seems like it has to be `>= min` or if it is really `max - 1` then it has to be `min  < max`

core/shared/src/test/scala/cats/parse/ParserTest.scala
johnynek: Looking at the new tests, I think these are good.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -990,12 +1005,23 @@ object Parser {
...
+    * note: this can wind up parsing nothing
+    */
+  def repAs0[A, B](p1: Parser[A], max: Int)(implicit acc: Accumulator0[A, B]): Parser0[B] =
+    Impl.Rep0(p1, max, acc)
 
   /** Repeat this parser 1 or more times
     */
   def repAs[A, B](p1: Parser[A], min: Int)(implicit acc: Accumulator[A, B]): Parser[B] =
-    Impl.Rep(p1, min, acc)
+    repAs(p1, min, Int.MaxValue)
```
johnynek, line 1019: Noting: here `Int.MaxValue` means infinity. But below, if a user passes Int.MaxValue they mean literally Int.MaxValue. I get your point that since Strings are indexed by Int, maybe this distinction is academic, but still I'd like to max the distinction just to make the code clearer.

core/shared/src/main/scala/cats/parse/Parser.scala
martijnhoekstra: This is here because we pre-parse one for the appender in rep (but not rep0 or the non-capturing varieties) before we do things with loops, so we need to parse at most max - 1 in the loop unless max == Int.MaxValue.

In the loop we test for `cnt <= loopmax - 1` so that if we need to enter the loop no more than `loopmax` times unless `loopmax == Int.MaxValue`, in which case we can enter the loop more than Int.MaxValue times because it never evaluates to false.

This wasn't apparent in the variation with the sentinel being a negative value, since it tested for max < 0, and it didn't matter whether max ended up being -1 or -2.

I'll see if I can find a clearer alternative encoding.

core/shared/src/main/scala/cats/parse/Parser.scala
martijnhoekstra: I think I understand your point now -- explicitly passed Int.MaxValue for max should never mean no maximum, but just mean Int.MaxValue times (even if the distinction at least for the time being is not observable), so the parsing methods that do the actual loop should represent their max value as max value - 1 to be able to distinguish between looping Int.MaxValue times and looping forever.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -990,12 +1005,23 @@ object Parser {
     * note: this can wind up parsing nothing
     */
   def repAs0[A, B](p1: Parser[A])(implicit acc: Accumulator0[A, B]): Parser0[B] =
-    Impl.Rep0(p1, acc)
+    Impl.Rep0(p1, Int.MaxValue, acc)
+
+  /** Repeat this parser 0 or more times
+    * note: this can wind up parsing nothing
+    */
+  def repAs0[A, B](p1: Parser[A], max: Int)(implicit acc: Accumulator0[A, B]): Parser0[B] =
```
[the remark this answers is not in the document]
martijnhoekstra: I'm having a hard time thinking of where allowing a Parser0 to repeat is actually what you want.

I can see `OneOf(P, P, P, P0).rep(max)`, which you should probably refactor to `OneOf(P, P, P).rep(max).orElse(P0)`, but maybe that's not always feasible.

I kind of to want the library to tell me I'm probably making a mistake, but when lemonade needs to be made, I don't want things to get in my way either. Introducing another semi-overload with another 0 somewhere -- rep0As0? -- is pretty gnarly too.

Which is long for 🤷

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -516,6 +520,11 @@ sealed abstract class Parser[+A] extends Parser0[A] {
   def rep(min: Int): Parser[NonEmptyList[A]] =
     Parser.repAs(this, min = min)
 
+  def rep(min: Int, max: Int): Parser[NonEmptyList[A]] = {
+    require(min >= 1, s"min should be >= 1 but was $min")
```
johnynek, line 524: Should we move this require into repAs? Seems this call has the require but the one above does not.

I tend to say yes.

core/shared/src/main/scala/cats/parse/Parser.scala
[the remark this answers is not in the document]
johnynek: I agree with your take.

core/shared/src/main/scala/cats/parse/Parser.scala
johnynek: Let's remove the commented code. If we are going to put assertions let's put them before parse time (at parser creation time).

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -2271,36 +2299,40 @@ object Parser {
...
+    case class Rep[A, B](p1: Parser[A], min: Int, maxMinusOne: Int, acc1: Accumulator[A, B])
+        extends Parser[B] {
       if (min < 1) throw new IllegalArgumentException(s"expected min >= 1, found: $min")
 
       private[this] val ignore: B = null.asInstanceOf[B]
 
       override def parseMut(state: State): B = {
         val head = p1.parseMut(state)
-
+        def maxRemaining = if (maxMinusOne == Int.MaxValue) Int.MaxValue else maxMinusOne - 1
```
johnynek, line 2328: Can we add a comment that this is to deal with the fact that we have already parsed one item, so if we are parsing a finite number we need to decrease by one.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -990,12 +1005,23 @@ object Parser {
...
   /** Repeat this parser 1 or more times
     */
   def repAs[A, B](p1: Parser[A], min: Int)(implicit acc: Accumulator[A, B]): Parser[B] =
-    Impl.Rep(p1, min, acc)
+    Impl.Rep(p1, min, Int.MaxValue, acc)
+
+  /** Repeat this parser 1 or more times
+    */
+  def repAs[A, B](p1: Parser[A], min: Int, max: Int)(implicit acc: Accumulator[A, B]): Parser[B] =
+    Impl.Rep(p1, min, max - 1, acc)
```
johnynek, line 1024: Either we should asset that min <= max or r we should return fail in the case max < min

What do you think? I tend towards error since that won't be recovered at parse time with a oneOf/orElse.

core/shared/src/main/scala/cats/parse/Parser.scala
martijnhoekstra: I'll take another pass over the require's and make sure everything in the layer that calls `Impl.Rep`, so that we're sure they're called exactly once, with min/max identical to the ones the user passed.

core/shared/src/main/scala/cats/parse/Parser.scala
martijnhoekstra: Let's go with the error at parser creation. If we return fail, there is the next sticky dilemma of whether it should consume up until `max` repetitions before failing or not. Just making it an error side-steps all that.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1086,6 +1109,126 @@ class ParserTest extends munit.ScalaCheckSuite {
...
+    forAll(ParserGen.gen, Gen.choose(0, Int.MaxValue), Arbitrary.arbitrary[String]) {
+      (genP, max, str) =>
+        genP.fa.rep0(0, max).parse(str).foreach { case (_, l) => assert(l.length <= max) }
+    }
+  }
+
+  property("repAs parses max entries when available") {
+    forAll(Gen.choose(1, 100)) { (n: Int) =>
+      val toBeConsumed = "a" * n
+      val p = Parser.char('a').repAs[Int](min = n, max = n)
```
johnynek, line 1163: can we exercise repNAs/exactlyAs/whatever here?

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1086,6 +1109,126 @@ class ParserTest extends munit.ScalaCheckSuite {
...
+      val p = Parser.anyChar.repAs0[Int](max = n)
+      val parsed = p.parse(toBeConsumed)
+      assertEquals(parsed, Right(("a", n)))
+    }
+  }
+
+  property("rep parses max entries when available") {
+    forAll(Gen.choose(1, 100)) { (n: Int) =>
+      val toBeConsumed = "a" * n
+      val p = Parser.char('a').rep(min = n, max = n).map(_.length)
```
johnynek, line 1199: can we use repN/exactly here?

core/shared/src/test/scala/cats/parse/RepParserConstructionTest.scala
```diff
@@ -0,0 +1,200 @@
...
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package cats.parse
+
+import org.scalacheck.Gen
+import org.scalacheck.Prop.forAll
+
+class RepParserConstructionTest extends munit.ScalaCheckSuite {
```
johnynek, line 27: these are nice tests.

johnynek (commented):
This is looking great.

One minor naming concern.

core/shared/src/main/scala/cats/parse/Parser.scala
martijnhoekstra: It's moved up. Would you still like to switch to `Long`?

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -987,15 +1003,55 @@ object Parser {
...
+    require(min >= 1, s"min should be >= 1, was $min")
+    require(max >= min, s"max should be >= min, but $max < $min")
+    Impl.Rep(p1, min, max - 1, acc)
+  }
+
+  /** Repeat the parser exactly `times` times
+    *
+    * @throws IllegalArgumentException if times < 1
+    */
+  def repNAs[A, B](p: Parser[A], times: Int)(implicit acc: Accumulator[A, B]): Parser0[B] = {
```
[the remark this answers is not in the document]
martijnhoekstra: repExactlyAs? At least for autocomplete it's nice to type `rep` and get the available options.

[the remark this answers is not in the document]
johnynek: sounds good.

core/shared/src/main/scala/cats/parse/Parser.scala
johnynek: I don't think we need that due to the point about String being limited to Int.MaxValue in length and this consuming at least one char per iteration.

core/shared/src/test/scala/cats/parse/ParserTest.scala
martijnhoekstra: The difference between this one and the one above if that one is `rep` and the other is `repAs`. If we use exactly for both, then they become identical.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1086,6 +1109,126 @@ class ParserTest extends munit.ScalaCheckSuite {
...
+  }
+
+  property("rep0 parses n entries, min <= n <= max") {
+    val validMinMax = for {
+      min <- Gen.choose(0, Int.MaxValue)
+      max <- Gen.choose(min, Int.MaxValue)
+    } yield (min, max)
+    forAll(ParserGen.gen, validMinMax, Arbitrary.arbitrary[String]) { (genP, minmax, str) =>
+      {
+        val (min, max) = minmax
```
martijnhoekstra, line 1144: This test failed to catch the max being discarded when min == 0 bug you discovered, even after running locally 10 times in a row with the bug active.

That suggests to me that this test (and the other ones that take arbitrary parsers and only may have meaningful results on success) only provide a false sense of security and a small contribution to global warming.

Shall we get rid of them? pretty much the same is tested in the rep tests on below with the "a" strings.

johnynek: I think it is fine to leave them because the law should be true. That said, I think the direct tests you added where the matches are more common are important to have as well.

Sometimes I've found bugs months after running property checks, so, I like having them.

johnynek (approved):

johnynek:
This is really great!

Thanks.

PR #133, opened 2021-01-03
Cleanup accumulated errors

regadas:
wdyt?

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -606,64 +606,58 @@ object Parser {
         }
       }
 
+    private def mergeInRange(irs: List[InRange]): List[InRange] = {
+      @annotation.tailrec
+      def merge(rs: List[InRange], aux: Chain[InRange] = Chain.empty): Chain[InRange] =
+        rs match {
+          case x :: y :: rest =>
+            if (y.lower.toInt > x.upper.toInt + 1) merge(y :: rest, aux :+ x)
+            else merge(InRange(x.offset, x.lower, x.upper max y.upper) :: rest, aux)
+          case _ =>
+            aux ++ Chain.fromSeq(rs.reverse)
+        }
+      merge(irs.sortBy(_.lower)).toList
+    }
+
+    private def mergeOneOfStr(ooss: List[OneOfStr]): Option[OneOfStr] =
+      if (ooss.isEmpty) None
+      else {
+        val ssb = SortedSet.newBuilder[String]
+        ooss.foreach(ssb ++= _.strs)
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -606,64 +606,58 @@ object Parser {
         }
       }
 
+    private def mergeInRange(irs: List[InRange]): List[InRange] = {
+      @annotation.tailrec
+      def merge(rs: List[InRange], aux: Chain[InRange] = Chain.empty): Chain[InRange] =
+        rs match {
+          case x :: y :: rest =>
+            if (y.lower.toInt > x.upper.toInt + 1) merge(y :: rest, aux :+ x)
```
johnynek, line 614: It's not clear to me this is merging the regions correctly. You aren't consulting y.upper.

Could we copy the old code out to a test and verify that the behavior is the same?

regadas: Sure thing. Added a test with it.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1556,7 +1550,7 @@ object Parser {
     var capture: Boolean = true
   }
 
-  private object Impl {
+  private[parse] object Impl {
```
regadas, line 1553: Made this `private[parse]` again so we can use `rangesFor` in tests

johnynek: rather than opening all of Impl, can we lift rangesFor out of Impl and make it `private[parse]`? Maybe irrationally, I prefer to use private as much as possible. I don't recall all the details, but some of the `private[foo]` stuff is actually public in the byte code.

Lastly, I think it can change inlining, but I'm not 100% sure about that.

regadas: I see your point. The whole access modifier representation in JVM is interesting. 

> private[foo] stuff is actually public in the byte code.

True, and the funny fact is that `private class / object` gets represented as `public`. the access modifier is only represented in attributes under `InnerClasses`. Methods however they do get their `private` modifier (expect the `private[foo]`). This might be even more relevant if Java interop is in mind.

Anyway initially, I was wondering if down the line we will need to tap into `Impl` a bit more for testing but I guess for now moving this one out is fine as well.

regadas:
From a couple of test runs, they do seem somewhat faster now. I reckon there's a lot of fail parses.

johnynek (approved):

johnynek:
Yeah, I think we almost always have failures honestly, since we are randomly generating parsers and sending random strings

Issue #137, opened 2021-01-04
publish 0.3.0


johnynek:
There is conceivably one last item to consider: #128 

Any others?

cc @regadas @martijnhoekstra

regadas:
iirc there were some ideas for additional combinators floating around, but I guess those can easily go on the next one.

For this one how are we with scalafix rules? it would be good if we could have them here in the repo. 
@martijnhoekstra did you start working on them? I can help out if you want.

martijnhoekstra:
The scalafix is ready to go. I'll give it another testrun on http4s.

I agree it's kind of weird to have them in my repo, rather than here, but also not really a big deal. Scalafix assumes `master` for the main branch if you use a `github:` scheme, so if we do it here, we need to use the raw path rather than a git path.

martijnhoekstra:
Other than that sounds good to me to release 0.3.0

martijnhoekstra:
Looking at the results in http4s, I still want to make two changes to the scalafix to make it catch more:

* Also rename methods that were imported rather than selected from the Parser object. in the current version P.char1(foo) gets replaced with P.char(foo), but import P.char1; char1(foo) remains unchanged. This pattern is heavily used in http4s
* Insert a pair of parentheses when translating from `Parser.rep1(foo)` to `(foo).rep` if foo is some expression that is not just a name. In practice, foo is often `x *> y`, and we want to get `(x *> y).rep`, not `x *> y.rep`

martijnhoekstra:
It's probably better to release now rip the Parser/Parser0 bandaid off as soon as possible. #128 needs to experiment and iterate a bit first, and I don't think it's worth holding a release up over.

As an example for the scalafix, https://github.com/http4s/http4s/compare/dotty...martijnhoekstra:rescaffoldFix?expand=1 has two commits, one with what the scalafix did, and one with the manual cleanup pass over it.

regadas:
Oh nice! Awesome to see that there are some initial rules. Having them on the repo would probably just facilitate contribution/awareness but not concerned if they are not.

I'm +1 on getting it out now.

martijnhoekstra:
This is where the rules are at atm @regadas  https://github.com/martijnhoekstra/catsparsescalafix/

johnynek:
okay, do you all think we are ready to tag 0.3.0?

johnynek:
maybe we should do #111 before we publish or close it. We have done so much rep/repSep related work on this version.

[...]

PR #139, opened 2021-01-05
Some more pre-0.3.0 changes

johnynek:
relates to #137 

This does a few things to make the API more consistent:

1. rename rep0Sep to repSep0. All the other methods with 0, such as repAs0, end in 0, this was the only one in the middle.
2. add a repSep variant with a maximum.
3. a few internal nits (use repExactly if possible, if repExactly is times = 1, just use map).

cc @regadas @martijnhoekstra

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -401,7 +401,7 @@ sealed abstract class Parser[+A] extends Parser0[A] {
   override def backtrack: Parser[A] =
     Parser.backtrack(this)
 
-  /** This method overrides `Parser#eitherOr` to refine the return type.
+  /** a version of eitherOr when both sides are not Parser0
     */
   def eitherOr[B](pb: Parser[B]): Parser[Either[B, A]] =
     Parser.eitherOr(this, pb)
@@ -647,11 +647,15 @@ object Parser {
           val fails = ListBuffer.empty[Fail]
           val others = ListBuffer.empty[Expectation]
 
-          list.toList.foreach {
-            case ir: InRange => rm += ir
-            case os: OneOfStr => om += os
-            case fail: Fail => fails += fail
-            case other => others += other
+          var items = list.toList
+          while (items.nonEmpty) {
+            items.head match {
+              case ir: InRange => rm += ir
+              case os: OneOfStr => om += os
+              case fail: Fail => fails += fail
+              case other => others += other
+            }
+            items = items.tail
           }
 
           // merge all the ranges:
@@ -805,7 +809,7 @@ object Parser {
       Parser.repAs(self, min = min, max = max)(acc)
 
     def repExactlyAs[B](times: Int)(implicit acc: Accumulator[A, B]): Parser[B] =
-      Parser.repAs(self, min = times, max = times)(acc)
+      Parser.repExactlyAs(self, times = times)(acc)
   }
 
   /** Don't advance in the parsed string, just return a
@@ -1035,17 +1039,23 @@ object Parser {
   def repAs[A, B](p1: Parser[A], min: Int, max: Int)(implicit acc: Accumulator[A, B]): Parser[B] = {
     require(min >= 1, s"min should be >= 1, was $min")
     require(max >= min, s"max should be >= min, but $max < $min")
-    Impl.Rep(p1, min, max - 1, acc)
+
+    if (min == max) repExactlyAs(p1, min)
+    else Impl.Rep(p1, min, max - 1, acc)
   }
 
   /** Repeat the parser exactly `times` times
     *
     * @throws java.lang.IllegalArgumentException if times < 1
     */
-  def repExactlyAs[A, B](p: Parser[A], times: Int)(implicit acc: Accumulator[A, B]): Parser0[B] = {
-    require(times >= 1, s"times should be >= 1, was $times")
-    Impl.Rep(p, times, times - 1, acc)
-  }
+  def repExactlyAs[A, B](p: Parser[A], times: Int)(implicit acc: Accumulator[A, B]): Parser[B] =
+    if (times == 1) {
+      // a map can be removed by a subsequent .void or .string
+      p.map { a => acc.newAppender(a).finish() }
+    } else {
+      require(times > 1, s"times should be >= 1, was $times")
+      Impl.Rep(p, times, times - 1, acc)
+    }
 
   /** Repeat 1 or more times with a separator
     */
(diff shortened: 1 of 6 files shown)


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1058,14 +1068,39 @@ object Parser {
 
   /** Repeat 0 or more times with a separator
     */
-  def rep0Sep[A](p1: Parser[A], min: Int, sep: Parser0[Any]): Parser0[List[A]] = {
+  def repSep0[A](p1: Parser[A], min: Int, sep: Parser0[Any]): Parser0[List[A]] = {
```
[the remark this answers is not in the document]
martijnhoekstra: Yes, we should

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1058,14 +1068,39 @@ object Parser {
...
+
+    val rest = (sep.void.with1.soft *> p1).rep0(min = min - 1, max = max - 1)
+    (p1 ~ rest).map { case (h, t) => NonEmptyList(h, t) }
+  }
+
+  /** Repeat 0 or more times with a separator
+    */
+  def repSep0[A](p1: Parser[A], min: Int, max: Int, sep: Parser0[Any]): Parser0[List[A]] = {
+    if (min < 0) throw new IllegalArgumentException(s"require min >= 0, found: $min")
+
```
martijnhoekstra, line 1093: no validation for max?

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1005,13 +1009,24 @@ object Parser {
 
   /** Repeat this parser 0 or more times, but no more than `max`
     *
-    * @throws java.lang.IllegalArgumentException if max < 1
+    * It may seem weird to accept 0 here, but without, composing
```
johnynek, line 1012: we call these functions from each other in a pretty confusing pattern. Once I was biasing the repetition count to small numbers, the tests were failing because we couldn't handle 0 here. I figured users could get confused too in complex calling situations and I figured 0 should be a valid value here.

Open to reconsidering.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -177,10 +177,24 @@ object ParserGen {
     GenT[Parser0, List[g.A]](g.fa.rep0(min = min, max = max))
   }
 
+  // this makes an integer >= min, but with power law bias to smaller values
+  def biasSmall(min: Int): Gen[Int] = {
```
johnynek, line 181: this turns out to really help us find some bugs. Previously, using choose was making it so we almost never selected 0 or 1, which are the common cases for users. This has the effect of making strong bias to 0, but still quite a good possibility of a larger value too.

cc @non something like this would be good to have baked into scalacheck.Gen.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1229,20 +1249,37 @@ class ParserTest extends munit.ScalaCheckSuite {
...
-        val min1 = if (min < 1) 1 else min
-        val p2a = Parser.repSep(genP.fa, min = min1, sep = Parser.unit)
-        val p2b = genP.fa.rep(min = min1)
+      val min1 = if (min < 1) 1 else min
+      val p2a = Parser.repSep(genP.fa, min = min1, sep = Parser.unit)
+      val p2b = genP.fa.rep(min = min1)
 
-        assertEquals(p2a.parse(str), p2b.parse(str))
+      assertEquals(p2a.parse(str), p2b.parse(str))
+    } &&
```
johnynek, line 1271: note the `&&`, without this, the first test does not run.

forAll returns a value, not a side effect.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1341,17 +1378,17 @@ class ParserTest extends munit.ScalaCheckSuite {
       val pa: Parser[Any] = p1.fa
       val pb: Parser[Any] = p2.fa
       assertEquals(pa.orElse(pb), pa | pb)
-    }
+    } &&
```
johnynek, line 1381: these tests were not running. It is kind of okay since they were mostly compilation tests, but still...

we were discarding non-unit values. Not okay.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1093,19 +1119,13 @@ class ParserTest extends munit.ScalaCheckSuite {
   }
 
   property("rep0 is consistent with rep") {
-    forAll(ParserGen.gen, Gen.choose(0, Int.MaxValue), Arbitrary.arbitrary[String]) {
-      (genP, min0, str) =>
-        val min = min0 & Int.MaxValue
-        val repA = genP.fa.rep0(min)
-        val repB = genP.fa
-          .rep(min)
```
johnynek, line 1101: this was a bug in the test: passing 0 here will throw, and that was possible before.

The solution is to start at 1.

core/shared/src/main/scala/cats/parse/Parser.scala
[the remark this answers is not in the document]
johnynek: after this is merged, please update the #137 issue when the scalafix is updated so we can safely publish (since I think scala-steward will use that scala fix).

core/shared/src/main/scala/cats/parse/Parser.scala
regadas: hm, this indeed seems confusing as with this `p1` just get's "ignored" right?

If I'm seeing this right one of the main issues is due to `repSep`. i.e for `repSep(string(...), 1, 1, string(...)))`

```scala
      // rest would be ignored
      val rest = (sep.void.with1.soft *> p1).rep0(min = min - 1, max = max - 1)
      (p1 ~ rest).map { case (h, t) => NonEmptyList(h, t) }
```

could we optimize it for:

```scala
    if (max == 1) {
      p1.map(NonEmptyList.one(_))
    } else {
      val rest = (sep.void.with1.soft *> p1).rep0(min = min - 1, max = max - 1)
      (p1 ~ rest).map { case (h, t) => NonEmptyList(h, t) }
    }
```

this way I guess we can keep the requirement on max >= 1. tests would need to be updated to make sure max >= 1

core/shared/src/test/scala/cats/parse/ParserTest.scala
regadas: nice catch!

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -647,11 +647,15 @@ object Parser {
           val fails = ListBuffer.empty[Fail]
           val others = ListBuffer.empty[Expectation]
 
-          list.toList.foreach {
-            case ir: InRange => rm += ir
-            case os: OneOfStr => om += os
-            case fail: Fail => fails += fail
-            case other => others += other
+          var items = list.toList
+          while (items.nonEmpty) {
```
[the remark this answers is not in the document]
martijnhoekstra: This parser will report its errors faster than all the competition 🥇

core/shared/src/main/scala/cats/parse/Parser.scala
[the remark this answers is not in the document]
martijnhoekstra: The scalafix is updated

core/shared/src/test/scala/cats/parse/ParserTest.scala
martijnhoekstra: Not to throw shade at scalac, but the C# compiler would have given me a warning here.

Okay, that's totally to throw shade at scalac.

core/shared/src/main/scala/cats/parse/Parser.scala
martijnhoekstra: If we do accept 0 here, we probably should for all `rep0` variations to be at least consistent.

johnynek: We do accept 0 everywhere that is a Parser0. We don't accept for Parser, since it has to consume something.

johnynek: @regadas we can't actually optimize it when the max == 1, because that means 0 or 1. We could use oneOf for that, but I doubt it will actually be faster.

regadas: @johnynek hm, maybe i'm sewing it wrong but the optimization would be on `def repSep[A](p1: Parser[A], min: Int, max: Int, sep: Parser0[Any]): Parser[NonEmptyList[A]]` which repeats 1 or more. 

The main idea here is not really the optimization but it will avoid calling `rep0(0, 0)` L1090

johnynek:
Updated. I still allow max == 0 just because we can without failing and matching all expectations. I would argue in FP we usually don't go overboard with exceptions on invalid input and try to be generous in returning something if it can lawfully be done. I think returning an empty Accumulator result value on max == 0 for a Parser0 seems fine.

regadas (approved):

PR #141, opened 2021-01-08
RepSep overload with no `min` parameter

### LINKED ISSUE #111: No default min value on rep(1)sep
Repsep is 0 or more and rep1sep is 1 or more, but you still have to pass in a minimum, which feels unnatural, especially for repsep where the minimum is implied to be 0 anyway. Default values of 0 or 1, or overloads to the same effect would be useful.

martijnhoekstra:
Closes #111 

The minimal implementation: does not include other possible extensions of repSep like repSepAs with accumulator or repSep with trailing separator.

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -1077,7 +1077,7 @@ object Parser {
 
   /** Repeat `min` or more times with a separator, at least once.
     *
-    * @throws IllegalArgumentException if `min <= 0`
+    * @throws java.lang.IllegalArgumentException if `min <= 0`
     */
   def repSep[A](p1: Parser[A], min: Int, sep: Parser0[Any]): Parser[NonEmptyList[A]] = {
     // we validate here so the message matches what the user passes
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1072,6 +1072,13 @@ object Parser {
 
   /** Repeat 1 or more times with a separator
     */
+  def repSep[A](p1: Parser[A], sep: Parser0[Any]): Parser[NonEmptyList[A]] =
+    repSep(p1, min = 1, sep)
+
+  /** Repeat `min` or more times with a separator, at least once.
+    *
+    * @throws IllegalArgumentException if `min <= 0`
```
regadas, line 1080: 🎨 unfortunately scaladoc will not like this. It will need to be `java.lang.IllegalArgumentException`

regadas (approved):
LGTM

core/shared/src/main/scala/cats/parse/Parser.scala
martijnhoekstra: fixed

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1072,6 +1072,13 @@ object Parser {
 
   /** Repeat 1 or more times with a separator
     */
+  def repSep[A](p1: Parser[A], sep: Parser0[Any]): Parser[NonEmptyList[A]] =
```
johnynek, line 1075: Can we add a variant for repSep0 where the min = 0?

johnynek (approved):

PR #142, opened 2021-01-08
More repSep related clean ups

johnynek:
part of #111 follow on to #141

cc @martijnhoekstra

### DIFF
--- a/core/shared/src/main/scala/cats/parse/SemVer.scala
+++ b/core/shared/src/main/scala/cats/parse/SemVer.scala
@@ -52,12 +52,12 @@ object SemVer {
 
   val preReleaseIdentifier: Parser[String] = alphanumericIdentifier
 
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/cats/parse/SemVer.scala
```diff
@@ -52,12 +52,12 @@ object SemVer {
 
   val preReleaseIdentifier: Parser[String] = alphanumericIdentifier
 
-  val dotSeparatedBuildIdentifiers: Parser[String] = Parser.repSep(buildIdentifier, dot).string
+  val dotSeparatedBuildIdentifiers: Parser[String] = buildIdentifier.repSep(dot).string
```
martijnhoekstra, line 55: These clean up really nice I think

martijnhoekstra (approved):
The methods on Parser really proof themselves here

PR #144, opened 2021-01-09
fix tests: use whole prefix length in case of repeating strings

martijnhoekstra:
These tests were failing if the input strings were repeating, and the part before the ~ matched the first half.


johnynek:
Good catch!

PR #145, opened 2021-01-09
remove nonNegativeDigit in SemVer

martijnhoekstra:
It's unused and there is no reason why you should use SemVer.positiveDigit over Numbers.nonZeroDigit

Could perhaps still go along with 0.3.0 in the hopes nobody uses it, and never will? Or deprecate first?


johnynek:
I'd be really shocked if anyone was using that.

PR #146, opened 2021-01-10
Add repeat Parser until end Parser

regadas:


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1446,12 +1456,23 @@ object Parser {
...
+  /** parse zero or more times until Parser `end` succeeds.
+    */
+  def until0[A](p: Parser[A], end: Parser[Any]): Parser0[List[A]] = {
+    (not(end).void.with1 ~ p).rep0.map(_.map(_._2))
+  }
+
+  /** parse one or more times until Parser `end` succeeds.
+    */
+  def until[A](p: Parser[A], end: Parser[Any]): Parser[NonEmptyList[A]] =
+    (not(end).void.with1 ~ p).rep.map(_.map(_._2))
```
johnynek, line 1475: Can we use *> here so we don't need the map? Also we don't need the void in left then.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1446,12 +1456,23 @@ object Parser {
...
   /** parse one or more characters as long as they don't match p
     */
   def until(p: Parser0[Any]): Parser[String] =
-    (not(p).with1 ~ anyChar).rep.string
+    (not(p).void.with1 ~ anyChar).rep.string
+
+  /** parse zero or more times until Parser `end` succeeds.
+    */
+  def until0[A](p: Parser[A], end: Parser[Any]): Parser0[List[A]] = {
+    (not(end).void.with1 ~ p).rep0.map(_.map(_._2))
```
johnynek, line 1469: Can we use *> so we don't need the map? We can also remove the void on the left since *> will do that.

johnynek (commented):
What do you think of renaming these method repUntil? That will help with autocomplete and emphasize that there is a repetition going on.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1446,12 +1456,23 @@ object Parser {
     *  this is useful for parsing comment strings, for instance.
     */
   def until0(p: Parser0[Any]): Parser0[String] =
-    (not(p).with1 ~ anyChar).rep0.string
+    (not(p).void.with1 ~ anyChar).rep0.string
```
johnynek, line 1459: you don't need this BTW, the `.string` would internally call .void.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -561,6 +561,16 @@ sealed abstract class Parser[+A] extends Parser0[A] {
   def repSep(min: Int, max: Int, sep: Parser0[Any]): Parser[NonEmptyList[A]] =
     Parser.repSep(this, min = min, max = max, sep = sep)
 
+  /** Repeat this parser 0 or more times until `end` Parser succeeds.
+    */
+  def until0(end: Parser[Any]): Parser0[List[A]] =
```
johnynek, line 566: actually, what about `def repUntil0[B](end: Parser[Any])(implicit acc: Accumulator0[A, B]): Parser0[B]`?

similar for `def repUntil` with `Accumulator`?

core/shared/src/main/scala/cats/parse/Parser.scala
regadas: 👍 right, it makes cleaner

core/shared/src/main/scala/cats/parse/Parser.scala
regadas: good point!

core/shared/src/main/scala/cats/parse/Parser.scala
regadas: yea, I forgot about them; we should add them.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -561,6 +561,26 @@ sealed abstract class Parser[+A] extends Parser0[A] {
   def repSep(min: Int, max: Int, sep: Parser0[Any]): Parser[NonEmptyList[A]] =
     Parser.repSep(this, min = min, max = max, sep = sep)
 
+  /** Repeat this parser 0 or more times until `end` Parser succeeds.
+    */
+  def repUntil0(end: Parser[Any]): Parser0[List[A]] =
```
johnynek, line 566: One minor issue: I think end can always be a Parser0 without sacrificing any safety.

For instance it could be a Parser.end which could be potentially useful or it could be a parser that would expects the index to be in a certain range.

johnynek (commented):
Looks great except one minor issue.

core/shared/src/main/scala/cats/parse/Parser.scala
regadas: Yea, I see, that does make sense. Let me update it.

johnynek (approved):
Thanks!

PR #148, opened 2021-01-10
Remove ParserMethods which is unneeded

johnynek:
Originally the `ParserMethods` class was intended to avoid writing a type like:

```scala
def repAs[A1 >: A, B](implicit acc: Accumulator[A1, B]): Parser[B]
```

But that is not needed because `Accumulator` and `Accumulator0` are contra variant in the first type parameter. Thus, we can put them as regular methods on `Parser[A]` without any ugly signatures and still be able to write `p.repAs[NonEmptyVector[Int]]` without having to name the input type (`Int` in this example).

This should be source compatible since we just moved implicit enrichments directly onto the class.

cc @martijnhoekstra @regadas

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -501,7 +501,7 @@ sealed abstract class Parser[+A] extends Parser0[A] {
   /** Repeat the parser `min` or more times, but no more than `max`
     *
     * The parser fails if it can't match at least `min` times
-    * After repeating the parser `max` times, the parser completes succesfully
+    * After repeating the parser `max` times, the parser completes successfully
     *
     * @throws java.lang.IllegalArgumentException if min < 0 or max < min
     */
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -573,7 +573,7 @@ sealed abstract class Parser[+A] extends Parser0[A] {
   /** Repeat the parser `min` or more times, but no more than `max`
     *
     * The parser fails if it can't match at least `min` times
-    * After repeating the parser `max` times, the parser completes succesfully
+    * After repeating the parser `max` times, the parser completes successfully
     *
     * @throws java.lang.IllegalArgumentException if min < 1 or max < min
     */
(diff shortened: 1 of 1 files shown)


martijnhoekstra:
Nice

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -497,31 +496,96 @@ sealed abstract class Parser[+A] extends Parser0[A] {
...
     * values are produced an arresting failure will be returned.
     */
   def rep(min: Int): Parser[NonEmptyList[A]] =
-    Parser.repAs(this, min = min)
+    repAs(min = min)
 
+  /** Repeat the parser `min` or more times, but no more than `max`
+    *
+    * The parser fails if it can't match at least `min` times
+    * After repeating the parser `max` times, the parser completes succesfully
```
regadas, line 531:
```suggestion
    * After repeating the parser `max` times, the parser completes successfully
```

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -497,31 +496,96 @@ sealed abstract class Parser[+A] extends Parser0[A] {
...
+    *
+    * @throws java.lang.IllegalArgumentException if min < 1
+    */
+  def repAs[B](min: Int)(implicit acc: Accumulator[A, B]): Parser[B] =
+    Parser.repAs(this, min = min)(acc)
+
+  /** Repeat the parser `min` or more times, but no more than `max`
+    *
+    * The parser fails if it can't match at least `min` times
+    * After repeating the parser `max` times, the parser completes succesfully
```
regadas, line 576:
```suggestion
    * After repeating the parser `max` times, the parser completes successfully
```

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -497,31 +496,96 @@ sealed abstract class Parser[+A] extends Parser0[A] {
     */
   def rep0(min: Int): Parser0[List[A]] =
     if (min == 0) rep0
-    else this.repAs(min)
+    else repAs(min)
 
+  /** Repeat the parser `min` or more times, but no more than `max`
+    *
+    * The parser fails if it can't match at least `min` times
+    * After repeating the parser `max` times, the parser completes succesfully
```
regadas, line 504:
```suggestion
    * After repeating the parser `max` times, the parser completes successfully
```

regadas (approved):

regadas:
@johnynek Hope you don't mind; they were just small typos.

PR #149, opened 2021-01-11
Add a recursive method

johnynek:
Add a helper method to make it cleaner to write recursive parsers.

cc @regadas @martijnhoekstra


regadas (approved):
Looks great!

PR #180, opened 2021-03-07
Update versions in README

ankitson:
Hi,

I just thought it might be useful to bump the versions in the README to the latest release version (0.3.1). `0.2.0` doesn't support Dotty so its inconsistent with the rest of the doc.


rossabaker (approved):
:+1: Thanks!

Issue #188, opened 2021-03-27
A | B != B | A


timzaak:
maybe, remove this `|` method.  just use `orElse`

johnynek:
While it is true that | is not commutative (a.backtrack | b.backtrack would be), having an infix operator is very nice.

Do you have any other suggestions?

timzaak:
maybe `|` would act like `oneOf`.

I just write graphql schema parser.
[Parser2.scala](https://github.com/timzaak/caliban/blob/cat-parse/core/src/main/scala/caliban/parsing/Parser2.scala). 
there are some not good user experience to share:
1. need insert lots of whitespaces, maybe it's a good idea to create and ParserContext, just like fastparse to handle whitespaces. 
2. so much `backtrack`,  and `?` always need `backtrack`. maybe it's my code fault, there would have nice methods to avoid.
3. `P[A] ~ (P[B].surround(whitespaces).backtrack.? ~  P[C].surround(whitespaces).backtrack.?)`  will read whitespaces twice when P[B] is failed. but I don't know how to improve it.

ritschwumm:
PEGs use `/` for their non-commutative OR - would that fit?

johnynek:
Sorry I'm slow to reply on the weekend.

You definitely don't always need backtrack with ?. I'll rewrite some of your examples without backtrack and it might give you a feel for how to do it.

A big design property of this library is to encourage you to write parsers avoiding backtracking since that can easily result in accidental exponential complexity and also it generally results in poor descriptions of what went wrong.

That said, we could probably do with better educational materials to help people get started.

johnynek:
Also note | does act like oneOf. It is a synonym. oneOf is also not commutative, it matches left to right.

timzaak:
@johnynek  thanks. And I'm sorry that I misunderstand the usage of `|`. It's good.

ritschwumm:
intuitively, i'd expect `|` to be commutative, too and therefore rename it to `/`. or is this a problem concerning operator precedence?

johnynek:
related to #184

johnynek:
While `/` is a good suggestion, the cost of breaking compatibility is significant... I can imagine adding `/` as another synonym of `orElse` but then we have three equivalent methods: orElse, `/` and `|`.

I'd like to take it a bit slow.

johnynek:
[...]

Issue #184, opened 2021-03-20
Parser.oneOf could use expanded documentation of when order matters


morgen-peschke:
It appears that order matters for `Parser.oneOf`, in cases where one parser accepts a subset of another parser. 

This makes sense, however it might be good to explicitly mention in the docs the implication this has for generating parsers for a set of `String` values - specifically that they should be reverse sorted according to length, because it's very easy to create inconsistent parsers if the input isn't correctly prepared.

For example, parsing a truthy value for `true` will consistently work (or fail) depending on the order of the parsers:
```scala
import cats.parse.{Parser => P}

val buggy: P[Boolean] =
  P.oneOf(List("1", "t", "tru", "yes", "true").map(P.string(_)))
    .void
    .as[Boolean](true)

val works: P[Boolean] =
  P.oneOf(List("true", "yes", "tru", "t", "1").map(P.string(_)))
    .void
    .as[Boolean](true)

def sort(strings: List[String]): List[String] =
  strings
    .map(str => (str.length, str))
    .sorted
    .reverse
    .map(_._2)

val sorted: P[Boolean] =
  P.oneOf(sort(List("1", "t", "tru", "yes", "true")).map(P.string(_)))
    .void
    .as[Boolean](true)

List("1", "t", "tru", "true", "y")
  .foreach { input =>
    println {
      """|%-6s => %6s => %s
         |%-6s => %6s => %s
         |%-6s => %6s => %s
         |""".stripMargin.format(
           s"<$input>", "buggy", buggy.parseAll(input),
           "", "works", works.parseAll(input),
           "", "sorted", sorted.parseAll(input)
         )
    }
  }
```

This is particularly troublesome because the error looks like an unexpected end of string, rather than an expected end of string that didn't happen:
```
Left(Error(1,NonEmptyList(EndOfString(1,3))))
```
[Scastie](https://scastie.scala-lang.org/OP8AVkBlROGYyvue66wM5w)

johnynek:
I'm actually don't see the error here.

oneOf should always care about ordering. It tries to match the left first, then moves to the right.

There are optimizations to improve performance, which could have bugs, but the design is to to match the left first then the right. So, if you match `t` or `true` the true will never match because `t` has already matched.

Can you write a unit test that you think should pass that does not?

The reason you see end of string, if you try to match `true` against `tru`, then you will hit the end of the string and have an error with that.

morgen-peschke:
[...]

PR #193, opened 2021-04-01
Add some comments to improve documentation

johnynek:
This adds some comments related to #184 and #188

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -211,6 +211,8 @@ sealed abstract class Parser0[+A] {
     Parser.oneOf0(this :: that :: Nil)
 
   /** Synonym for orElse
+    *  Note this is not commutative: if this has an arresting failure we
+    *  do not continue onto the next.
     */
   def |[A1 >: A](that: Parser0[A1]): Parser0[A1] =
     orElse(that)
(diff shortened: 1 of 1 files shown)


johnynek:
@morgen-peschke and @timzaak what do you all think of these minor documentation improvements?

rossabaker (approved):

kubukoz (approved):

timzaak:
LGTM

nafg:
Where is "arresting failure" defined?

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -962,6 +966,11 @@ object Parser {
     *  see @backtrack if you want to do backtracking.
     *
     *  This is the same as parsers.foldLeft(fail)(_.orElse(_))
+    *
+    *  recommended style: oneOf(p1 :: p2 :: p3 :: Nil) rather than
```
nafg, line 970: Why do you care how people define their Lists?

johnynek: Using List(a, b, c) uses a var args constructor which allocates and then builds a list after. Using this style avoids that.

IMO it looks nicer to not have the List(..) present as well.

That is why I recommend it.

johnynek:
arresting failure is defined in the comments on the Parser.scala file:

https://github.com/typelevel/cats-parse/blob/65c79bddfd300dd13717525dbc6ba05d5fa70dda/core/shared/src/main/scala/cats/parse/Parser.scala#L45

morgen-peschke:
Looks good 👍

PR #194, opened 2021-04-02
Add withContext to improve error reporting

### LINKED ISSUE #43: design for scoping parsers for error reporting
Currently, when a parser fails you get a nonempty list of offsets and failed expectations.

We could also add "scope" wrapper, like: `number.scope("number").orElse(str.scope("string"))`. How this would work is the mutable State would have a stack of these scopes, and we have:

```scala
case class ScopeParser[A](parser: Parser[A], scope: String) extends Parser[A]
```
in parseMut we would push the current scope onto the stack, then parse with parser, then pop it off the stack.

When we error, we take a snapshot of the current scope stack.

So, if we do this, users can have an easier time labeling parts of their parsers and seeing where things went wrong.

Fastparse does something similar.

What do you think of this design @mpilquist @non @rossabaker and really anyone who cares to comment.

johnynek:
close #43 

This follows the design sketched in the issue above. The goals are:

1. not break compatibility
2. not add a performance tax on those not using the feature
3. enable users to add strings to track where they are in a parser.

cc @kubukoz @mpilquist

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -354,6 +354,12 @@ sealed abstract class Parser0[+A] {
   def surroundedBy(b: Parser0[Any]): Parser0[A] =
     between(b, b)
 
+  /** Add a string context to any Errors on parsing
+    *  this is useful for debugging failing parsers.
+    */
+  def withContext(str: String): Parser0[A] =
+    Parser.withContext0(this, str)
+
   /** Internal (mutable) parsing method.
     *
     * This method should only be called internally by parser instances.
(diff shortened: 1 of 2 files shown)


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -672,6 +685,16 @@ object Parser {
     */
   sealed abstract class Expectation {
     def offset: Int
+
+    /** This is a reverse order stack (most recent context first)
+      * of this parsing error
+      */
+    def context: List[String] =
```
kubukoz, line 692: Can we make this `tailrec`? It'd probably require a `reverse` at the end but maybe it's worth pursuing?

kubukoz (approved):
Looks good to me! Although some test with an example and not as much logic as in the existing prop tests would be nice.

core/shared/src/main/scala/cats/parse/Parser.scala
johnynek: Seems to me if your errors are so nested you blow the stack here you will have other issues. I don't think it is worth a cost to retain the ability of having 50k deep nested contexts.

What do you think?

kubukoz: Fair. Let's leave it ;)

johnynek:
This is published as part of 0.3.3

PR #200, opened 2021-04-11
Cache hashCode

### LINKED ISSUE #198: CPU time spent in hashcode computation
Hi,
We are currently trying to replace fastparse by cats-parse and ran into the issue that parsing became 3 to 30x slower than it was with fastparse. Using a profiler, I saw that most of the CPU is actually spent on calculating hashcode, triggered by the use of `.distinct` in `oneOf`. See the following screenshots:

Screen Shot 2021-04-10 at 3 07 25 PM
Screen Shot 2021-04-10 at 3 09 53 PM

Do you think this is caused by a wrong use of cat-parse, or is it something that can improved in cats-parse itself? Parser code can be found [here](https://github.com/ghostdogpr/caliban/pull/834/files#diff-03436de4a1f55d01b6f3ca37b998f2c4293c8b0249fa1d1fc9c6afbe22538dfb) if it helps.

ghostdogpr:
Resolves #198

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -58,7 +58,7 @@ import java.util.Arrays
   * Rewinding tends to make error reporting more difficult and can lead
   * to exponential parser behavior it is not the default behavior.
   */
-sealed abstract class Parser0[+A] {
+sealed abstract class Parser0[+A] extends Product {
 
   /** Attempt to parse an `A` value out of `str`.
     *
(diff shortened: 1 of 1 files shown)


ghostdogpr:
Looks like it's not binary compatible.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -58,7 +58,7 @@ import java.util.Arrays
   * Rewinding tends to make error reporting more difficult and can lead
   * to exponential parser behavior it is not the default behavior.
   */
-sealed abstract class Parser0[+A] {
+sealed abstract class Parser0[+A] extends Product {
```
johnynek, line 61: Can we do `{ self: Product =>` in abstract class? I forget. That may allow us to bypass the binary compact problem since all subclasses are Products.

ghostdogpr: Seems like it works, if I add it to `Parser` as well.

regadas (approved):
Lgtm 👍

johnynek (approved):
Thanks!

johnynek:
This is published as part of 0.3.3

PR #210, opened 2021-04-30
Fix sbt 1.5.1 deprecated scoping

regadas:


johnynek:
Thanks!

PR #222, opened 2021-05-14
Upgrade to Scala 3

mpilquist:


johnynek:
great!

if you push a tag according to this:

https://github.com/typelevel/cats-parse/blob/main/Release.md

it will be published. Seems like 0.3.4 would be the version number.

Nice to have 2.11 to 3.0 support!

:)

mpilquist:
Awesome, just tagged v0.3.4

Issue #234, opened 2021-06-03
API to read source position


re-xyr:
As title. ANTLR allows to read the source interval (start offset, end offset, start row, end row, start col, end col) of any node. Trifecta has `DeltaParsing`. I did not find such an API here and I think this is a crucial feature for any further error reporting after parsing.

johnynek:
There is Parser.index which gives the current offset, which can be used to get the range: get index, parse, get the index. If you want line and column numbers see LocationMap which can convert offsets into rows and columns.

re-xyr:
I see. Do we want to provide `Caret`s and `Span`s like Trifecta? If so maybe I can make a pr.

johnynek:
Yeah, that could be nice, especially if we can do it without breaking compatibility.

PR #238, opened 2021-06-12
add caret and spans like Trifecta (#234)

re-xyr:
Solves #234; adds `Parser(0)#careted, spanned, spanning` and `Parser.careted`.

I'm unsure about the performance of these combinators but I expect it to be not very good. This is due to recomputation in `LocationMap` each time the combinators are used. If this is much of a concern, this PR may be incapable. A probable optimization is to store the current `Position` (incl. row, col) in the `State` and update by need when the combinators are called (this is the approach of Megaparsec).


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1731,6 +1764,57 @@ object Parser {
     */
   def end: Parser0[Unit] = Impl.EndParser
 
+  /** return the input string we are parsing.
+    */
+  def input: Parser0[String] = Impl.Input
```
johnynek, line 1769: I'm a bit nervous about exposing this.

It closes off the possibility of ever easily supporting Streaming.

You could imagine parsing `Stream[String]` or `LazyList[String]`, but if you have people reaching for the whole String, that is going to blow up their memory.

On the other hand, I guess a streaming parser is just really different from a parser that parses a whole file. A streaming parser should be something like: `Stream[F, String] => Stream[ParserF[F, *], A]` or something...

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1731,6 +1764,57 @@ object Parser {
...
+
+  /** return the position the parser is at
+    */
+  def careting: Parser0[Caret] =
+    (input, index).tupled map { case (in, ix) =>
+      val locmap = LocationMap(in)
```
johnynek, line 1775: what about making `careting` the fundamental thing instead of using input, then adding to `State` a `lazy val locationMap: LocationMap = LocationMap(str)`:

https://github.com/typelevel/cats-parse/blob/8c9a27143abc30bc7f8000feb25d75e9a3a10906/core/shared/src/main/scala/cats/parse/Parser.scala#L1880

then you only compute the LocationMap once per parse and only if needed by a parser?

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1731,6 +1764,57 @@ object Parser {
...
   def end: Parser0[Unit] = Impl.EndParser
 
+  /** return the input string we are parsing.
+    */
+  def input: Parser0[String] = Impl.Input
+
+  /** return the position the parser is at
+    */
+  def careting: Parser0[Caret] =
+    (input, index).tupled map { case (in, ix) =>
```
johnynek, line 1774: please avoid infix use of methods (deprecated in scala 3).

Can we instead do `(input, index).mapN { (in, ix) => ...`

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -668,6 +668,18 @@ class ParserTest extends munit.ScalaCheckSuite {
     parseTest(Parser.stringIn(List("foo", "foobar", "foofoo", "foobat")).string, "foobal", "foo")
   }
```
johnynek, line 670: please add generators that create any new elements to the AST (in this PR that would be `input`, but in a follow up it might be `WithCaret` or something). That way we can check all the other properties in the presence of this new node.

johnynek:
Thanks for sending this PR! I think we will find a way to merge something close to this.

core/shared/src/main/scala/cats/parse/Parser.scala
re-xyr: I see. In Megaparsec they used a lazily updated state field that stores source position and a snapshot of the stream.

core/shared/src/main/scala/cats/parse/Parser.scala
re-xyr: Oh sorry, I only read the Scala 2 style guide and saw that the infix form *was* recommended...

re-xyr:
I am thinking about `Span`. Do you think `(Caret, Caret)` will be better than a separate datatype?

core/shared/src/main/scala/cats/parse/Parser.scala
re-xyr: Thanks! Now I have `Impl.WithCaret` instead and a `val locmap` in `State`.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1731,6 +1764,46 @@ object Parser {
...
+  /** return the result of the parser together
+    * with the position of the starting of the consumption
+    */
+  def careted[A](pa: Parser[A]): Parser[(A, Caret)] =
+    (careting.with1 ~ pa).map { case (l, x) => (x, l) }
+
+  /** replace the result of the parser with with the
+    * span of position of the string it consumed
+    */
+  def spanning0[A](pa: Parser0[A]): Parser0[Span] =
```
johnynek, line 1786: can we change this to 'def spanOf`?

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1731,6 +1764,46 @@ object Parser {
...
+  /** replace the result of the parser with with the
+    * span of position of the string it consumed
+    */
+  def spanning[A](pa: Parser[A]): Parser[Span] =
+    (careting.with1 ~ (pa *> careting)).map { case (l, r) => Span(l, r) }
+
+  /** return the result of the parser together
+    * with the span of the string it consumed
+    */
+  def spanned0[A](pa: Parser0[A]): Parser0[(A, Span)] =
```
johnynek, line 1798: can we call this `def withSpan`.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1878,6 +1951,7 @@ object Parser {
    * This is protected rather than private to avoid a warning on 2.12
    */
   protected[parse] final class State(val str: String) {
+    lazy val locmap: LocationMap = LocationMap(str)
```
johnynek, line 1954: nit: can we rename `locMap`

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -2209,6 +2284,15 @@ object Parser {
       override def parseMut(state: State): Int = state.offset
     }
 
+    case object WithCaret extends Parser0[Caret] {
+      override def parseMut(state: State): Caret = {
+        val (row, col) = state.locmap
+          .toLineCol(state.offset)
+          .getOrElse(throw new IllegalStateException("the offset is larger than the input size"))
```
johnynek, line 2291: this will throw when you are at the EOF right?

I think we need to fix that somehow.

I guess we can say that EOF should return `toLineCol(eof - 1).map { case (row, col) => (row, col + 1) }`

What do you think?

core/shared/src/test/scala/cats/parse/ParserTest.scala
johnynek: I still want to address this to make sure all the previous laws pass.

see for instance:

https://github.com/typelevel/cats-parse/blob/8c9a27143abc30bc7f8000feb25d75e9a3a10906/core/shared/src/test/scala/cats/parse/ParserTest.scala#L167

as an example of how we want to transform some generators to make `withSpan` and `spanOf` and add them to here:

https://github.com/typelevel/cats-parse/blob/8c9a27143abc30bc7f8000feb25d75e9a3a10906/core/shared/src/test/scala/cats/parse/ParserTest.scala#L527

and:

https://github.com/typelevel/cats-parse/blob/8c9a27143abc30bc7f8000feb25d75e9a3a10906/core/shared/src/test/scala/cats/parse/ParserTest.scala#L555

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -668,6 +668,18 @@ class ParserTest extends munit.ScalaCheckSuite {
...
 
+  test("Parser.careted and spanned works") {
```
johnynek, line 671: also, I'd like to see a law like `pa.parse(str) == pa.withSpan.map(_._1).parse(str)` and `pa.spanOf.parse(str) == pa.withSpan.map(_._2).parse(str)`

something like that.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1731,6 +1764,46 @@ object Parser {
...
   def end: Parser0[Unit] = Impl.EndParser
 
+  /** return the position the parser is at
+    */
+  def careting: Parser0[Caret] = Impl.WithCaret
+
+  /** return the result of the parser together
+    * with the position of the starting of the consumption
+    */
+  def careted0[A](pa: Parser0[A]): Parser0[(A, Caret)] =
```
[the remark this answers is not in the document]
re-xyr: My idea was that `pa` was primary and the caret was secondary so the result of `pa` comes first, but yes that would cause confusion about where is the caret exactly. If you prefer putting the caret at the first element I'll change it.

I think `pa.careted` is more readable than `(caret ~ pa)` in the sense that it associates tighter in sight which indicates it is somehow more of a single parsing unit. Indeed we can leave this out and let the user make the choice whether to write this helper instead since it is really short.

core/shared/src/main/scala/cats/parse/Parser.scala
re-xyr: I think `span` is better given we have `string`?

core/shared/src/main/scala/cats/parse/Parser.scala
re-xyr: Yes. I'll take a look at how I can solve that first.

core/shared/src/main/scala/cats/parse/Parser.scala
johnynek: Sure, that's fine.

core/shared/src/main/scala/cats/parse/Parser.scala
re-xyr: Added `GenT` for spanning combinators but not yet added to the list in `gen(0)`. Will add after this is solved.

core/shared/src/main/scala/cats/parse/LocationMap.scala
```diff
@@ -91,3 +93,13 @@ class LocationMap(val input: String) {
 object LocationMap {
   def apply(str: String): LocationMap = new LocationMap(str)
 }
+
+case class Caret(offset: Int, row: Int, col: Int)
+object Caret {
+  implicit val eqCatsCaret: Eq[Caret] = Eq.fromUniversalEquals
```
johnynek, line 99: what about making an `Order` instead and order by the `offset`? We might want to sort by carets.

core/shared/src/main/scala/cats/parse/LocationMap.scala
```diff
@@ -91,3 +93,13 @@ class LocationMap(val input: String) {
...
+}
+
+case class Span(from: Caret, to: Caret)
+object Span {
+  implicit val eqCatsCaret: Eq[Span] = Eq.fromUniversalEquals
```
johnynek, line 104: also we could make an `Order` here where we compare first by from, then to.

core/shared/src/main/scala/cats/parse/LocationMap.scala
```diff
@@ -91,3 +93,13 @@ class LocationMap(val input: String) {
 object LocationMap {
   def apply(str: String): LocationMap = new LocationMap(str)
 }
+
+case class Caret(offset: Int, row: Int, col: Int)
+object Caret {
+  implicit val eqCatsCaret: Eq[Caret] = Eq.fromUniversalEquals
+}
+
+case class Span(from: Caret, to: Caret)
```
johnynek, line 102: can we add a comment if from and to are inclusive/exclusive? I assume `to` is exclusive.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -360,6 +360,24 @@ sealed abstract class Parser0[+A] { self: Product =>
   def withContext(str: String): Parser0[A] =
     Parser.withContext0(this, str)
 
+  /** return the result of the parser together
+    * with the position of the starting of the consumption
+    */
+  def withCaret: Parser0[(A, Caret)] =
```
johnynek, line 366: what about call this `def caretBefore: Parser0[(Caret, A)]` or something?

I'm really nervous `withCaret` and the fact that the caret is on the right is going to suggest to people that the caret is after the parse.

core/shared/src/test/scala/cats/parse/ParserTest.scala
johnynek: I see that you added the generators, but I don't see that they were added to the root generator so we don't exercise all the existing laws in the presence of these operations.

johnynek:
do you think you will be able to address the comments I left?

re-xyr:
Sorry, I think I've been busy in August. I may be able to address them in September.

core/shared/src/main/scala/cats/parse/Parser.scala
johnynek: #286 fixes this issue.

johnynek:
I'm planning to publish a new version soon if you have time to pick this back up.

PR #62, opened 2020-11-13
Small optimization around makeError

regadas:
before:
```
[info] Benchmark                           Mode  Cnt    Score    Error  Units
[info] BarBench.catsParseParse             avgt    3   ≈ 10⁻³           ms/op
[info] Bla25Bench.catsParseParse           avgt    3   42.306 ±  6.775  ms/op
[info] CountriesBench.catsParseParse       avgt    3   16.722 ±  4.179  ms/op
[info] Qux2Bench.catsParseParse            avgt    3   13.183 ±  1.485  ms/op
[info] Ugh10kBench.catsParseParse          avgt    3  100.625 ± 27.054  ms/op
```

after:
```
[info] Benchmark                           Mode  Cnt   Score    Error  Units
[info] BarBench.catsParseParse             avgt    3  ≈ 10⁻⁴           ms/op
[info] Bla25Bench.catsParseParse           avgt    3  35.326 ±  8.988  ms/op
[info] CountriesBench.catsParseParse       avgt    3  14.109 ±  3.304  ms/op
[info] Qux2Bench.catsParseParse            avgt    3  10.504 ±  2.995  ms/op
[info] Ugh10kBench.catsParseParse          avgt    3  88.645 ± 29.553  ms/op
```

PR #280, opened 2021-10-15
Defer creation of Expectations

rossabaker:
Given a parser whose hot part looks like this:

```scala
    val fieldName = Rfc7230.token.string
    val fieldValue = P.repSep0(Rfc5234.vchar.rep, P.charIn(" \t").rep).string.surroundedBy(Rfc7230.ows)
    val header = ((fieldName <* P.char(':')) ~ fieldValue).map {
      case (k, v) => Header.Raw(CIString(k), v)
    }
    val headers = (header <* P.string("\r\n")).rep0
```

`CharIn.makeError` is dominating the stack trace and allocations.  It is created at the end of each repetition, only to be nulled out and gc'ed if we've made any progress.

```
[info] Benchmark                                                         Mode  Cnt       Score       Error   Units
[info] Http1DecoderBench.http1Decoder                                   thrpt    5  194717.904 ± 27142.805   ops/s
[info] Http1DecoderBench.http1Decoder:·gc.alloc.rate                    thrpt    5    2375.264 ±   330.882  MB/sec
[info] Http1DecoderBench.http1Decoder:·gc.alloc.rate.norm               thrpt    5   13436.503 ±     0.524    B/op
[info] Http1DecoderBench.http1Decoder:·gc.churn.G1_Eden_Space           thrpt    5    2378.683 ±   323.143  MB/sec
[info] Http1DecoderBench.http1Decoder:·gc.churn.G1_Eden_Space.norm      thrpt    5   13456.252 ±   205.497    B/op
[info] Http1DecoderBench.http1Decoder:·gc.churn.G1_Survivor_Space       thrpt    5       0.008 ±     0.004  MB/sec
[info] Http1DecoderBench.http1Decoder:·gc.churn.G1_Survivor_Space.norm  thrpt    5       0.044 ±     0.019    B/op
[info] Http1DecoderBench.http1Decoder:·gc.count                         thrpt    5     508.000              counts
[info] Http1DecoderBench.http1Decoder:·gc.time                          thrpt    5     552.000                  ms
[info] Http1DecoderBench.http1Decoder:·stack                            thrpt              NaN                 ---
```

Here, we wrap every `Chain[Expectation]` in an `Eval`, so that the expectations can be created lazily.  Care must be taken not to close over mutable state, namely, `state.offset`.  We significantly increase speed and reduce garbage:

```
[info] Http1DecoderBench.http1Decoder                                   thrpt    5  353576.391 ± 39077.640   ops/s
[info] Http1DecoderBench.http1Decoder:·gc.alloc.rate                    thrpt    5    2293.983 ±   253.014  MB/sec
[info] Http1DecoderBench.http1Decoder:·gc.alloc.rate.norm               thrpt    5    7146.434 ±     0.264    B/op
[info] Http1DecoderBench.http1Decoder:·gc.churn.G1_Eden_Space           thrpt    5    2294.376 ±   247.756  MB/sec
[info] Http1DecoderBench.http1Decoder:·gc.churn.G1_Eden_Space.norm      thrpt    5    7147.791 ±    94.745    B/op
[info] Http1DecoderBench.http1Decoder:·gc.churn.G1_Survivor_Space       thrpt    5       0.007 ±     0.002  MB/sec
[info] Http1DecoderBench.http1Decoder:·gc.churn.G1_Survivor_Space.norm  thrpt    5       0.023 ±     0.006    B/op
[info] Http1DecoderBench.http1Decoder:·gc.count                         thrpt    5     494.000              counts
[info] Http1DecoderBench.http1Decoder:·gc.time                          thrpt    5     519.000                  ms
[info] Http1DecoderBench.http1Decoder:·stack                            thrpt              NaN                 ---
```

I'll work on cleaning up and providing a runnable benchmark, but wanted to get early feedback.

We might also make it leaner with simple thunks rather than an `Eval` that's always `later`.

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -68,7 +68,7 @@ sealed abstract class Parser0[+A] { self: Product =>
     val offset = state.offset
     if (err eq null) Right((str.substring(offset), result))
     else
-      Left(Parser.Error(offset, Parser.Expectation.unify(NonEmptyList.fromListUnsafe(err.toList))))
+      Left(Parser.Error(offset, Parser.Expectation.unify(NonEmptyList.fromListUnsafe(err.value.toList))))
   }
 
   /** Attempt to parse all of the input `str` into an `A` value.
@@ -92,7 +92,7 @@ sealed abstract class Parser0[+A] { self: Product =>
           )
         )
     } else
-      Left(Parser.Error(offset, Parser.Expectation.unify(NonEmptyList.fromListUnsafe(err.toList))))
+      Left(Parser.Error(offset, Parser.Expectation.unify(NonEmptyList.fromListUnsafe(err.value.toList))))
   }
 
   /** Convert epsilon failures into None values.
@@ -1773,7 +1773,7 @@ object Parser {
    */
   protected[parse] final class State(val str: String) {
     var offset: Int = 0
-    var error: Chain[Expectation] = null
+    var error: Eval[Chain[Expectation]] = null
     var capture: Boolean = true
   }
 
@@ -2033,7 +2033,7 @@ object Parser {
           state.offset = end
           res
         } else {
-          state.error = Chain.one(Expectation.Length(offset, len, state.str.length - offset))
+          state.error = Eval.later(Chain.one(Expectation.Length(offset, len, state.str.length - offset)))
           null
         }
       }
@@ -2079,8 +2079,9 @@ object Parser {
 
     case object StartParser extends Parser0[Unit] {
       override def parseMut(state: State): Unit = {
-        if (state.offset != 0) {
-          state.error = Chain.one(Expectation.StartOfString(state.offset))
+        val offset = state.offset
+        if (offset != 0) {
+          state.error = Eval.later(Chain.one(Expectation.StartOfString(offset)))
         }
         ()
       }
@@ -2088,8 +2089,9 @@ object Parser {
 
     case object EndParser extends Parser0[Unit] {
       override def parseMut(state: State): Unit = {
-        if (state.offset != state.str.length) {
-          state.error = Chain.one(Expectation.EndOfString(state.offset, state.str.length))
+        val offset = state.offset
+        if (offset != state.str.length) {
+          state.error = Eval.later(Chain.one(Expectation.EndOfString(offset, state.str.length)))
         }
         ()
       }
@@ -2128,7 +2130,7 @@ object Parser {
           state.offset += message.length
           ()
         } else {
-          state.error = Chain.one(Expectation.OneOfStr(offset, message :: Nil))
+          state.error = Eval.later(Chain.one(Expectation.OneOfStr(offset, message :: Nil)))
           ()
         }
       }
@@ -2144,7 +2146,7 @@ object Parser {
           state.offset += message.length
           ()
         } else {
-          state.error = Chain.one(Expectation.OneOfStr(offset, message :: Nil))
+          state.error = Eval.later(Chain.one(Expectation.OneOfStr(offset, message :: Nil)))
           ()
         }
       }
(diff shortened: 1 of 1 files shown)


johnynek:
The win looks really good and I'm definitely interested in merging when green and we sort out any mima issues.

rossabaker:
Drat.  I was afraid stack safety would be an issue with the thunks.  I'll pause here and see if people agree with the problem before I pour more into the solution.

rossabaker:
A thunk is slightly lighter than `Eval`, but not much more so, and Eval does indeed take care of stack safety.

[Replacing `Eval.later` with `Eval.always`](https://github.com/rossabaker/cats-parse/commit/098d02000d427f4dfa62578bbfdf3828deb58f89) is marginally slower, but eliminates some allocations:

```
[info] Http1DecoderBench.http1Decoder                                   thrpt    5  337942.154 ± 42747.443   ops/s
[info] Http1DecoderBench.http1Decoder:·gc.alloc.rate                    thrpt    5    2072.398 ±   261.164  MB/sec
[info] Http1DecoderBench.http1Decoder:·gc.alloc.rate.norm               thrpt    5    6754.582 ±     0.565    B/op
[info] Http1DecoderBench.http1Decoder:·gc.churn.G1_Eden_Space           thrpt    5    2075.145 ±   264.493  MB/sec
[info] Http1DecoderBench.http1Decoder:·gc.churn.G1_Eden_Space.norm      thrpt    5    6763.527 ±   113.341    B/op
[info] Http1DecoderBench.http1Decoder:·gc.churn.G1_Survivor_Space       thrpt    5       0.008 ±     0.003  MB/sec
[info] Http1DecoderBench.http1Decoder:·gc.churn.G1_Survivor_Space.norm  thrpt    5       0.025 ±     0.009    B/op
[info] Http1DecoderBench.http1Decoder:·gc.count                         thrpt    5     480.000              counts
[info] Http1DecoderBench.http1Decoder:·gc.time                          thrpt    5     520.000                  ms
[info] Http1DecoderBench.http1Decoder:·stack                            thrpt              NaN                 ---
```

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -2656,15 +2672,14 @@ object Parser {
...
-          aux = aux.tail
-        }
-        result
+      def makeError(offset: Int): Eval[Chain[Expectation]] = {
+        Eval.later(
+          Chain.fromSeq(
+            ranges.toList.map { case (s, e) =>
+              Expectation.InRange(offset, s, e)
+            }
+          )
```
regadas, line 2681: @rossabaker last time I saw this https://github.com/typelevel/cats-parse/pull/62 are you seeing diff results? maybe worth keep?

regadas (approved):
This looks great! those numbers look good 😄

rossabaker:
I need my CPU back so I can't run a proper number of iterations right now, but looking good on the JSON:

`bench/jmh:run -i 4 -wi 2 -f 1 -t 1 .*catsParseParse`

Before, on 9e67814:

```
[info] Benchmark                   Mode  Cnt   Score    Error  Units
[info] BarBench.catsParseParse     avgt    4  ≈ 10⁻⁴           ms/op
[info] Bla25Bench.catsParseParse   avgt    4  36.105 ±  4.108  ms/op
[info] Qux2Bench.catsParseParse    avgt    4  10.878 ±  1.018  ms/op
[info] Ugh10kBench.catsParseParse  avgt    4  84.775 ±  4.723  ms/op
```

After, on 9561249:

```
[info] Benchmark                   Mode  Cnt   Score    Error  Units
[info] BarBench.catsParseParse     avgt    4  ≈ 10⁻⁴           ms/op
[info] Bla25Bench.catsParseParse   avgt    4  30.502 ±  1.736  ms/op
[info] Qux2Bench.catsParseParse    avgt    4   9.649 ±  1.508  ms/op
[info] Ugh10kBench.catsParseParse  avgt    4  76.212 ±  6.961  ms/op
```

Error on both:

```
[info] java.lang.RuntimeException: Error(256941,NonEmptyList(InRange(256941,,,,)))
[info] 	at scala.sys.package$.error(package.scala:27)
[info] 	at cats.parse.bench.JmhBenchmarks.catsParseParse(JsonBench.scala:47)
[info] 	at cats.parse.bench.jmh_generated.CountriesBench_catsParseParse_jmhTest.catsParseParse_avgt_jmhStub(CountriesBench_catsParseParse_jmhTest.java:190)
[info] 	at cats.parse.bench.jmh_generated.CountriesBench_catsParseParse_jmhTest.catsParseParse_AverageTime(CountriesBench_catsParseParse_jmhTest.java:152)
```

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -2144,29 +2157,31 @@ object Parser {
...
+        val offset = state.offset
+        state.error = Eval.later(Chain.one(Expectation.FailWith(offset, message)))
         null.asInstanceOf[A]
       }
     }
 
     final def oneOf[A](all: Array[Parser0[A]], state: State): A = {
       val offset = state.offset
-      var errs: Chain[Expectation] = Chain.nil
+      var errs: Eval[Chain[Expectation]] = Eval.later(Chain.nil)
```
johnynek, line 2184: can we make this `Eval.now(Chain.nil)` I think that will be less allocation and actually every `oneOf` hits this path.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -2180,7 +2195,7 @@ object Parser {
           // we failed to parse, but didn't consume input
           // is unchanged we continue
           // else we stop
-          errs = errs ++ err
+          errs = errs.map(_ ++ err.value)
```
johnynek, line 2198: `.value` isn't safe. We should do `for { e1 <- errs; e2 <- err } yield (e1 ++ e2)` or `(errs, err).mapN(_ ++ _)` but maybe the latter is slightly slower due to dispatch via typeclass.

Issue #239, opened 2021-06-13
Parser error prettyprinting


re-xyr:
As title, it is good if we can format the `Expectation`s into `String`s. Ideally, Trifecta's output style (with line numbers and offending lines) will be very nice.

![](https://i.imgur.com/ZyzUSSn.png)

johnynek:
Yes. See the LocationMap class that can compute the line and column.

It would be nice to have some utility that allows you to easily print.

I have some in another project but I use typelevel/paiges to produce the result and I didn't want to add that dependency here.

But we could do something with String or StringBuilder perhaps.

re-xyr:
I wrote a simple prettyprinting utility: [gist link](https://gist.github.com/re-xyr/b0f1988e744c93d4c83a5f17f58484ea).

I tried to test with the JSON parser in README. It gave these errors:

```
(repl):2:7: error: expected one of 'false', 'null', 'true'
2 | {"x": except}
  |       ^
(repl):2:7: error: expected '"'
2 | {"x": except}
  |       ^
(repl):2:7: error: expected '0' ~ '9'
2 | {"x": except}
  |       ^
(repl):2:7: error: expected '['
2 | {"x": except}
  |       ^
(repl):2:7: error: expected '{'
2 | {"x": except}
  |       ^
```

Should these errors (all at one position) be combined in some way?

re-xyr:
Also, `LocationMap` does not support getting the EOF position (`offset == length`) which is inconvenient.

johnynek:
Yeah, that is an issue.

If you get `None` for the line/col, it might be EOL. It would have been nicer if I had noticed that earlier and I could have made an ADT return type:

```scala
sealed abstract class RowColResult {
  def toValid: Option[(Int, Int)]
}

object RowColResult {
  case object OutOfBounds extends RowColResult {
    def toValid = None
  }
  case class Valid(row: Int, col: Int) extends RowColResult {
    def toValid = Some((row, col))
  }
  case object EOF extends RowColResult {
    def toValid = None
  }
}
```
something like that...

We could probably add a method like that, and then have the current method just call `toValid` on that.

Want to make a PR?

johnynek:
About this: https://github.com/typelevel/cats-parse/issues/239#issuecomment-860181108

yes, I would group by location, and then aggregate all the expecations:
`expected false, true, null, ", [, or {`

something like that.

re-xyr:
[...]

PR #286, opened 2021-10-29
Make LocationMap return a line/col on EOF

johnynek:
address the issue here:

https://github.com/typelevel/cats-parse/issues/239#issuecomment-860196982

cc @re-xyr @alissa-tung please take a look at review if you have time.


regadas (approved):
Yeah, this makes sense! 👍 

> It's always a bit dangerous to change semantics without the types, but I lean towards accepting it here.

I agree

PR #290, opened 2021-10-30
Add soft.with1.between and surroundedBy

johnynek:
this was overlooked in #289


regadas (approved):
Right! nice catch @johnynek!

PR #291, opened 2021-10-30
Improve merging of Char parsers

johnynek:
1. If we have a single CharIn parser, we don't need to compute the union, just reuse it (which also saves recomputing the range of characters)
2. return a Set in the union function, not a List, which can return a large list if you have duplicates in your set.

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -2622,7 +2622,7 @@ object Parser {
             // and any direct prefix CharIns
             val tail1 = tail.filterNot(_.isInstanceOf[CharIn])
             (result :+ AnyChar.asInstanceOf[P0]) ++ Chain.fromSeq(tail1)
-          case (ci @ CharIn(_, _, _)) :: tail =>
+          case (ci: CharIn) :: tail =>
             loop(tail, ci :: front, result)
           case h :: tail =>
             // h is not an AnyChar or CharIn
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -2603,11 +2603,17 @@ object Parser {
      */
     def mergeCharIn[A, P0 <: Parser0[A]](ps: List[P0]): List[P0] = {
       @annotation.tailrec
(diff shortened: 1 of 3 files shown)


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -2616,8 +2622,8 @@ object Parser {
             // and any direct prefix CharIns
             val tail1 = tail.filterNot(_.isInstanceOf[CharIn])
             (result :+ AnyChar.asInstanceOf[P0]) ++ Chain.fromSeq(tail1)
-          case CharIn(m, bs, _) :: tail =>
-            loop(tail, (m, bs) :: front, result)
+          case (ci @ CharIn(_, _, _)) :: tail =>
```
regadas, line 2625:
```suggestion
          case (ci: CharIn) :: tail =>
```

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -2603,11 +2603,17 @@ object Parser {
...
         def frontRes: Chain[P0] =
-          if (front.isEmpty) Chain.nil
-          else Chain.one(Parser.charIn(BitSetUtil.union(front)).asInstanceOf[P0])
+          front match {
+            case Nil => Chain.nil
+            case one :: Nil => Chain.one(one.asInstanceOf[P0])
+            case many =>
+              // we need to union
+              val minBs: List[(Int, BitSetUtil.Tpe)] = many.map { case CharIn(m, bs, _) => (m, bs) }
+              Chain.one(Parser.charIn(BitSetUtil.union(minBs)).asInstanceOf[P0])
```
regadas, line 2615: could we have a version of `union` with `Iterator` and avoid the copy in `minBs`?

johnynek: we certainly could. I wasn't sure it was worth it because we already iterate through the number of parsers regularly, but the characters *inside* the parsers (which we have to iterate when allocating a new CharIn to compute the ranges) could be more expensive, which was what I was targeting here.

Do you want me to address this before merging?

regadas: We can address this as a follow-up for sure!

PR #295, opened 2021-11-06
Fix json code, run bench

johnynek:
Just wanted to update to see how much @rossabaker 's code in #280 helped the benchmarks (and fix an issue with the json parsing).


johnynek:
the gap with fast parse is pretty minimal now... we could possibly catch it!

johnynek:
one thing that isn't really fair... our string parsing code is actually doing decoding of unicode, where as the fast parse code isn't...

So, the logic is more complex.

rossabaker (approved):
Oh, geez.  Thank you for finishing #280.  I hadn't forgotten, but I hadn't made time... or yet noticed that you had.

PR #297, opened 2021-11-06
Make sure void is idempotent

johnynek:
This adds a law (that actually already passed), and makes unmap a bit more precise in the case of Defer.


regadas (approved):
Makes sense!

PR #298, opened 2021-11-08
Tutorial in readme

DenisNovac:
Hi everyone. Still not sure if i can handle the recursion topic though but here is the most of the library tutorial (i guess). 

Also something strange happens in diff section. It says index.md is binary. Is it ok?


johnynek:
Hey, thanks for the PR but it looks like maybe you forgot to add some files. I only see removed content, am I missing something?

DenisNovac:
I am confused too. I was editing index.md but git says it is binary. I can see the proper diff with `--text` flag:

```
$ git diff main change_readme --text

diff --git a/.gitignore b/.gitignore
index d55f078..7504387 100644
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,6 @@ TAGS
 .metals
 .bloop
 .bsp
+project/metals.sbt
+project/project/metals.sbt
+project/project/project/metals.sbt

diff --git a/docs/src/index.md b/docs/src/index.md
index 36eb3c6..7c7ea70 100644
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -27,7 +27,271 @@ this library has a few goals:
 5. Safety: by separating Parser0, a parser that may consume no input, from Parser, a parser must consume at least one character on success. Most combinators and methods can be made safer to use and less prone to runtime errors.
 6. Stability: we are very reluctant to break compatibility between versions. We want to put a minimal tax on users to stay on the latest versions.
 
-# An Example
+# Tutorial
+
+## Simple parser
+
+The library provides a set of simple parsers which might be combined to create any parsing logic. The simplest parser is `Parser.anyChar` which is successful where there is one char at the input. It has type `Parser[Char]` which means it returns one parsed char.

...

+So when the *right side* throws an error the `soft` method allows us to rewind parsed input and try to proceed it's parsing with next parsers (without changing the parser itself!).
+
+
+# JSON parser example
 Below is most of a json parser (the string unescaping is elided). This example can give you a feel
 for what it is like to use this library.
```

DenisNovac:
Oh i got it, i've inserted some EOL (or something) from the Metals console. I'll make another PR

PR #299, opened 2021-11-08
Add tutorial to readme

DenisNovac:
Fixed version of #298

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -35,7 +35,7 @@ The library provides a set of simple parsers which might be combined to create a
 
 To provide any input to parser one need to use `parse` method.
 
-```scala
+```scala mdoc:reset
 import cats.parse.Parser
 
 val p: Parser[Char] = Parser.anyChar
@@ -54,7 +54,7 @@ Notice the return type. `Tuple2[String, Char]` contains the rest of the input st
 
 The output of the parser might be processed with `map` method:
 
-```scala
+```scala mdoc:reset
 import cats.parse.Parser
 
 case class CharWrapper(value: Char) 
@@ -67,7 +67,7 @@ p.parse("t")
 
 There are built-in methods for mapping the output to types `String` or `Unit`:
 
-```scala
+```scala mdoc:reset
 import cats.parse.Rfc5234.digit
 import cats.parse.Parser
 
@@ -103,7 +103,7 @@ The parsers might be combined through operators:
 
 For this example we'll be using `cats.parse.Rfc5234` package which contains such parsers as `alpha` (Latin alphabet) and `sp` (whitespace). 
 
-```scala
+```scala mdoc:reset
 import cats.parse.Rfc5234.{sp, alpha, digit}
 import cats.parse.Parser
 
@@ -163,10 +163,10 @@ p3.parse(" ")
 
 Sometimes we need something to repeat zero or more types. The cats-parse have `rep` and `rep0` methods for repeating values. `rep` means that the parser must be successful *at least one time*. `rep0` means that the parser output might be empty.
 
-```scala
+```scala mdoc:reset
 import cats.data.NonEmptyList
 import cats.parse.Rfc5234.alpha
-import cats.parse.Parser
+import cats.parse.{Parser, Parser0}
 
 val p1: Parser[NonEmptyList[Char]]  = alpha.rep
 val p2: Parser0[List[Char]] = alpha.rep0
@@ -183,7 +183,7 @@ Notice the types of parsers. `Parser` type always means some non-empty output an
 
 One common task in this example is to parse a full line (or words) of text. In the example it is done by `rep`, and then it could be mapped to `String` in different ways:
 
-```scala
+```scala mdoc:reset
 import cats.data.NonEmptyList
 import cats.parse.Rfc5234.alpha
 import cats.parse.Parser
@@ -201,7 +201,7 @@ All three parsers will be identical in parsing results, but `p2` and `p3` are us
 
 Some parsers never return a value. They have a type `Parser0`. One might get this type of parser when using `rep0` or `.?` methods.
 
-```scala
+```scala mdoc:reset
 import cats.parse.Rfc5234.{alpha, sp}
 import cats.parse.Parser
 
@@ -213,7 +213,7 @@ p.parse("hello world")
 
 Notice the type we got - `Parser[String]`. That is because we have `rep` outside and our `alpha.rep` parser with `Parser` type is on the left side of the clause. But what if we want to parse strings with spaces at the beginning?
 
-```scala
+```scala:fail
 val p = (sp.? *> alpha.rep <* sp.?).rep.string
 ```
 
@@ -221,7 +221,7 @@ We will get an error `value rep is not a member of cats.parse.Parser0`. This hap
 
 But this parser can't be empty because of `alpha.rep` parser, and we know it. For these types of parsers we need to use `with1` wrapper method on the *left side* of the clause:
 
-```scala
+```scala mdoc:reset
 import cats.parse.Rfc5234.{alpha, sp}
 import cats.parse.Parser
 
@@ -243,7 +243,7 @@ Parser might be interrupted by parsing error. There are two kinds of errors:
  - an error that has consumed 0 characters (**epsilon failure**);
  - an error that has consumed 1 or more characters (**arresting failure**) (sometimes called halting failure).
 
-```scala
+```scala mdoc:reset
 import cats.parse.Rfc5234.{alpha, sp}
 import cats.parse.Parser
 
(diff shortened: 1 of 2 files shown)


zsluedem:
Just want to say . This pr is great.

docs/src/index.md
```diff
@@ -27,7 +27,271 @@ this library has a few goals:
...
+
+Notice the type we got - `Parser[String]`. That is because we have `rep` outside and our `alpha.rep` parser with `Parser` type is on the left side of the clause. But what if we want to parse strings with spaces at the beginning?
+
+```scala
+val p = (sp.? *> alpha.rep <* sp.?).rep.map { words => 
+  words.map(charsList => charsList.toList.mkString("")).toList.mkString(" ")
+}
+```
+
+We will get an error `value rep is not a member of cats.parse.Parser0`. This happens since we have the left-side parser as optional in `sp.? *> alpha.rep <* sp.?` clause. This clause has a type `Parser0` which can't be repeated.
```
johnynek, line 160: maybe we should not that we can do `alpha.repAs[String]` or `alpha.rep.string` (in this particular case, although not in general, they are the same). That is better than `rep` + `map` because we don't make the intermediate list in that case. We actually use a `StringBuilder` internally to build up the result.

docs/src/index.md
```diff
@@ -27,7 +27,271 @@ this library has a few goals:
...
+
+But this parser can't be empty because of `alpha.rep` parser, and we know it. For these types of parsers we need to use `with1` wrapper method on the *left side* of the clause.
+
+```scala
+import cats.parse.Rfc5234.{alpha, sp}
+import cats.parse.Parser
+
+
+val p: Parser[String] = (sp.?.with1 *> alpha.rep <* sp.?).rep.map { words =>
```
johnynek, line 169: should we mention the convenience method `.between` or `.surroundedBy`
```scala
(sp.?.with1 *> alpha.rep <* sp.?)  == alpha.rep.surroundedBy(sp.?)
```

docs/src/index.md
```diff
@@ -27,7 +27,271 @@ this library has a few goals:
...
+// res0: Either[Error, Tuple2[String, String]] = Right((,hello world))
+p.parse(" hello world")
+// res1: Either[Error, Tuple2[String, String]] = Right((,hello world))
+```
+
+If we have multiple `Parser0` parsers before the `Parser` - we'd need to use parenthesis like this:
+`(sp.? ~ sp.?).with1 *> alpha.rep`.
+
+## Error handling
+
```
johnynek, line 183: there are two kinds of errors:

1. an error that has consumed 0 characters (epsilon failure)
2. an error that has consumer 1 or more characters (halting failure).

I think it is important to mention that. `orElse` or `|` allows recovering from type 1, but not type 2 errors. `.backtrack` converts type 2 errors into type 1, but we want to generally minimize that for performance and locating errors.

docs/src/index.md
```diff
@@ -27,7 +27,271 @@ this library has a few goals:
...
+val p2 = fieldValueSoft.? ~ (searchWord ~ sp.?).rep.string
+
+p2.parse("title:The Wind Has Risen")
+// res2 = Right((,(Some((title,())),The Wind Has Risen)))
+p2.parse("The Wind Has Risen")
+// res3 = Right((,(None,The Wind Has Risen)))
+```
+
+So when the *right side* throws an error the `soft` method allows us to rewind parsed input and try to proceed it's parsing with next parsers (without changing the parser itself!).
```
johnynek, line 291: when the right side throws a *non-halting* error (which is to say epsilon error, or the right side did not consume any input).

johnynek (commented):
This is really great and I want to merge it!

I did have a few suggestions, mostly around calling out the two kinds of conceptual errors: halting and epsilon, and making the distinction early to help understand orElse, backtrack and soft.

regadas (commented):
This looks great! 👏 

Besides @johnynek comments can we make the code blocks `mdoc` blocks? It will make sure the examples compile and eval result will always be accurate.

docs/src/index.md
DenisNovac: In the backtrack docs there is a phrase "This method converts arresting failures into epsilon failures". Is arresting = halting?

johnynek: yeah, sorry. I forgot the terminology used in the docs (I think @non came up with that but I always seem to wind up recalling halting).

https://github.com/typelevel/cats-parse/blob/714a9df6ac458ba14d654ef4d5c620452db02cb3/core/shared/src/main/scala/cats/parse/Parser.scala#L43

Erik wrote some really nice docs, so let's stick to his terminology.

docs/src/index.md
```diff
@@ -181,9 +239,31 @@ If we have multiple `Parser0` parsers before the `Parser` - we'd need to use par
 
 ## Error handling
 
+Parser might be interrupted by parsing error. There are two kinds of errors:
+ - an error that has consumed 0 characters (**epsilon failure**);
+ - an error that has consumer 1 or more characters (**halting failure**).
```
johnynek, line 244: sorry for misleading you, let's stick to `arresting` used in the docs.

docs/src/index.md
```diff
@@ -181,9 +239,31 @@ If we have multiple `Parser0` parsers before the `Parser` - we'd need to use par
...
+p2.parse(" 1")
+// res1: Either[Error, Tuple2[String, Char]] = Left(Error(1,NonEmptyList(InRange(1,A,Z), InRange(1,a,z))))
+```
+
+We need to make this difference because the first type of error allows us to say that parser is not matching the input before we started to process it and the second error happens while parser processing the input.
+
 ### Backtrack
 
-Parser job might be interrupted by parsing error. This happens when input can't be parsed with this parser:
+Backtrack allows us to convert a *halting failure* to *epsilon failure*. It also rewinds the input to the offset to that used before parsing began. The resulting parser might still be combined with others. Let's look at the example:
```
johnynek, line 266: arresting

docs/src/index.md
```diff
@@ -234,13 +316,13 @@ val p3 = digit
 // res2 = Right((,1))
 ```
 
-The first parsing is interrupted by error since it *started* the parsing (one first space was correct on parser `p1`) so it won't work without `backtrack`. The second parsing works fine with `orElse` alone since it does not start in `p1` or `p2`.
+The first parser combination is interrupted by *halting error* and the second parsing combination will only suffer from *epsilon errors*. The second parser works because `orElse` and `|` operators actually allows recovering from epsilon errors, but not from halting errors.
```
johnynek, line 319: arresting.

docs/src/index.md
```diff
@@ -234,13 +316,13 @@ val p3 = digit
...
 
-So the `backtrack` helps us where the *left side* throws an error.
+So the `backtrack` helps us where the *left side* throws halting error.
```
johnynek, line 321: arresting

docs/src/index.md
```diff
@@ -234,13 +316,13 @@ val p3 = digit
...
 
 ### Soft
 
-This method might look similar to `backtrack`, but it allows us to *proceed* the parsing when the *right side* is failed. It is really useful for ambiguous parsers when we can't really tell what exactly we are parsing before the end. Let's say we want to parse some input to the search engine which contains fields. This might look like "field:search_query". Let's try to write a parser for this:
+This method might look similar to `backtrack`, but it allows us to *proceed* the parsing when the *right side* is throwing an epsilon error. It is really useful for ambiguous parsers when we can't really tell what exactly we are parsing before the end. Let's say we want to parse some input to the search engine which contains fields. This might look like "field:search_query". Let's try to write a parser for this:
```
johnynek, line 325: maybe `returning` than `throwing` which sounds like a jvm exception being thrown, which is not how it happens.

docs/src/index.md
```diff
@@ -273,7 +355,7 @@ val p3 = (searchWord ~ sp.?).rep.string
 // res1 = Right((,The Wind Has Risen))
 ```
 
-But this problem might be resolved with `soft` method:
+But this problem might be resolved with `soft` method inside the first parser since the right side of it actually throws an epsilon failure itself:
```
johnynek, line 358: `returns an epsilon`

johnynek (commented):
I agree with @regadas we should use `scala mdoc` so we compile these docs to make sure the examples compile.

See for instance here:
https://github.com/typelevel/cats-parse/blob/714a9df6ac458ba14d654ef4d5c620452db02cb3/docs/src/index.md?plain=1#L39

I am looking forward to publishing the next version of this library, but I'd love to merge this first so new users will see this tutorial if they see the announcement of a new version.

DenisNovac:
Fixed the terminology. I'll check mdoc later today.

DenisNovac:
Added the mdoc and checked it compiles

DenisNovac:
Fixed the warning (through web-version, didn't compile it for other warnings)

johnynek:
Thank you so much!

filosganga:
It is just what I was looking for, thank you!

PR #301, opened 2021-11-11
Add Parser.caret

johnynek:
This does the caret part of #238 

related to #234 

I did not add Span, just Caret.

My concern was that the span methods are a bit less clear. For instance, we may want an inclusive upper bound range, or maybe we do want exclusive. Inclusive is a bit more complex to implement however.

To sidestep this concern for now while we think about the design, I just added getting the `Caret` itself.

cc @re-xyr

### DIFF
--- a/core/shared/src/main/scala/cats/parse/LocationMap.scala
+++ b/core/shared/src/main/scala/cats/parse/LocationMap.scala
@@ -66,22 +66,29 @@ class LocationMap(val input: String) {
     */
   def toLineCol(offset: Int): Option[(Int, Int)] =
     if (offset < 0 || offset > input.length) None
+    else {
+      val Caret(_, row, col) = toCaretUnsafe(offset)
+      Some((row, col))
+    }
+
+  /** Convert an offset to a Caret.
+    * @throws IllegalArgumentException
+    *   if offset is longer than input
+    */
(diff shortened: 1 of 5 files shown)


johnynek:
actually, inclusive span could be bad because that means you can never use it with Parser0, but then again, maybe that is a feature? Maybe you only want the Span of a Parser?

core/shared/src/main/scala/cats/parse/LocationMap.scala
```diff
@@ -92,13 +99,17 @@ class LocationMap(val input: String) {
...
+        Caret(offset, row, col)
       } else {
         // idx is exactly the right value because offset is beginning of a line
-        Some((idx, 0))
+        Caret(offset, idx, 0)
       }
     }
 
+  def toCaret(offset: Int): Option[Caret] =
+    if (offset < 0 || offset > input.length) None
```
regadas, line 110: :art: this condition is popping a few times would a `isValidOffset` be worth?

regadas (approved):

core/shared/src/main/scala/cats/parse/LocationMap.scala
johnynek: good idea. I've made that change (and avoid checking it so many times).

PR #302, opened 2021-11-11
Remove mima exclusion

johnynek:
By putting the call to `makeError` inside `Eval.later` we can avoid the mima exclusion, which simplifies the build.

This is in preparation for publishing a new version hopefully today.


regadas (approved):
Nice 👍

PR #305, opened 2021-11-12
Fix some minor issues before 0.3.5

johnynek:
I realized we used `line` instead of `row` in the `LocationMap` and we had a doc error that caused `v0.3.5` publish to fail, so I figured I could fix that.


johnynek:
I am going to exercise a rare self merge so we can get this published today.

I think consistency with `LocationMap.toLineCol` means we should use `line` in `Caret`.

PR #308, opened 2021-11-12
Add BitSet union iterator

regadas:
Follow up to https://github.com/typelevel/cats-parse/pull/291/files#r740001716

Might not be a huge deal but it doesn't hurt.


johnynek (approved):
Thanks!

This could improve parser construction time (which is when this code runs).

Better to be more efficient in any case.

PR #313, opened 2021-11-20
fix Caret arguments ordering

hugo-vrijswijk:
`line` was filled with `offset`, `col` was filled with `line` and `offset` with `col`.

I'm guessing at one point Caret was `Caret(offset, line, column)`, but was changed to `Caret(line, column, offset)` without re-examing argument order.

This caused this behavior:

```scala
val p = (Parser.caret.with1 <* Parser.anyChar).rep

val pattern = """aaa"""

val parsed = p.parseAll(pattern).right.get
// parsed: cats.data.NonEmptyList[Caret] = NonEmptyList(Caret(0, 0, 0), List(Caret(1, 0, 1), Caret(2, 0, 2)))
```

Expected:
```scala
// parsed: cats.data.NonEmptyList[Caret] = NonEmptyList(Caret(0, 0, 0), List(Caret(0, 1, 1), Caret(0, 2, 2)))
```

### DIFF
--- a/core/shared/src/test/scala/cats/parse/LocationMapTest.scala
+++ b/core/shared/src/test/scala/cats/parse/LocationMapTest.scala
@@ -226,7 +226,7 @@ class LocationMapTest extends munit.ScalaCheckSuite {
         val lc = lm.toLineCol(offset)
 
         assertEquals(oc, Some(c))
-        assertEquals(lc, oc.map { case Caret(_, r, c) => (r, c) })
+        assertEquals(lc, oc.map { case c  => (c.line, c.col) })
       }
 
       if (other < 0 || s.length < other) {
--- a/core/shared/src/test/scala/cats/parse/LocationMapTest.scala
+++ b/core/shared/src/test/scala/cats/parse/LocationMapTest.scala
@@ -226,7 +226,7 @@ class LocationMapTest extends munit.ScalaCheckSuite {
         val lc = lm.toLineCol(offset)
 
         assertEquals(oc, Some(c))
-        assertEquals(lc, oc.map { case c  => (c.line, c.col) })
+        assertEquals(lc, oc.map { case c => (c.line, c.col) })
       }
 
       if (other < 0 || s.length < other) {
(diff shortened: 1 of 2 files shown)


core/shared/src/test/scala/cats/parse/LocationMapTest.scala
```diff
@@ -226,7 +226,7 @@ class LocationMapTest extends munit.ScalaCheckSuite {
         val lc = lm.toLineCol(offset)
 
         assertEquals(oc, Some(c))
-        assertEquals(lc, oc.map { case Caret(_, r, c) => (r, c) })
+        assertEquals(lc, oc.map { case c  => (c.line, c.col) })
```
johnynek, line 229: can we remove the `case` and the extra space here?

I think you want to run `fmt`.

hugo-vrijswijk: Yeah forgot to run formatting

core/shared/src/test/scala/cats/parse/LocationMapTest.scala
```diff
@@ -226,7 +226,7 @@ class LocationMapTest extends munit.ScalaCheckSuite {
         val lc = lm.toLineCol(offset)
 
         assertEquals(oc, Some(c))
-        assertEquals(lc, oc.map { case Caret(_, r, c) => (r, c) })
+        assertEquals(lc, oc.map { case c => (c.line, c.col) })
```
johnynek, line 229: reminder: can we remove the `case` now?

hugo-vrijswijk: Yes

johnynek (approved):
thanks! sorry I let this bug slip through.

core/shared/src/test/scala/cats/parse/LocationMapTest.scala
```diff
@@ -226,7 +226,7 @@ class LocationMapTest extends munit.ScalaCheckSuite {
         val lc = lm.toLineCol(offset)
```
johnynek, line 227: actually while we are at it can we add: `assertEquals(c.offset, offset)` to make sure we pass through the offset?

regadas:
dang it, I actually feel a bit responsible for this issue ... I should have caught it! Sorry about this!

johnynek:
No need to apologize!

PR #314, opened 2021-11-21
Bump Readme versions and add some tests

johnynek:
Does a few things:

1. bump the version mentioned in the readme
2. increase the number of test runs for LocationMapTest (which still doesn't it make it run more than a couple of seconds)
3. add a test that the pattern match on Caret is a certain order (the same order as comparison)
4. explain in a comment why the comparison order is what it is (in case we are comparing Carets from two different strings).

### DIFF
--- a/core/shared/src/test/scala/cats/parse/BitSetTest.scala
+++ b/core/shared/src/test/scala/cats/parse/BitSetTest.scala
@@ -25,7 +25,7 @@ import org.scalacheck.Prop.forAll
 
 class BitSetTest extends munit.ScalaCheckSuite {
   test("isScalaJs/isScalaJvm is consistent") {
-    // This will need to be updated is we ever add scala-native
+    // This will need to be updated if we ever add scala-native
     assert(!(BitSetUtil.isScalaJs && BitSetUtil.isScalaJvm))
     assert(BitSetUtil.isScalaJs || BitSetUtil.isScalaJvm)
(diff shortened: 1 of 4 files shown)


core/shared/src/test/scala/cats/parse/BitSetTest.scala
```diff
@@ -24,6 +24,13 @@ package cats.parse
 import org.scalacheck.Prop.forAll
 
 class BitSetTest extends munit.ScalaCheckSuite {
+  test("isScalaJs/isScalaJvm is consistent") {
+    // This will need to be updated is we ever add scala-native
```
regadas, line 28:
```suggestion
    // This will need to be updated if we ever add scala-native
```

core/shared/src/test/scala/cats/parse/LocationMapTest.scala
```diff
@@ -228,6 +246,10 @@ class LocationMapTest extends munit.ScalaCheckSuite {
         assertEquals(oc, Some(c))
         assertEquals(lc, oc.map { c => (c.line, c.col) })
         assertEquals(c.offset, offset)
+        val Caret(line, col, off1) = c
+        assertEquals(line, c.line)
+        assertEquals(col, c.col)
+        assertEquals(off1, offset)
```
regadas, line 252: 👍

PR #322, opened 2021-12-02
Update sbt-github-actions and sbt-spiewak-sonatype

regadas:


build.sbt
```diff
@@ -32,23 +32,23 @@ ThisBuild / githubWorkflowBuild := Seq(
   )
 )
 
+ThisBuild / githubWorkflowJavaVersions := List(JavaSpec.temurin("8"))
```
regadas, line 35: Thinking about this ... I believe we should probably just use `11` but still target `8`

johnynek (commented):
Wait so this means all users going forward not on 2.11 have to be on Java 11 right?

I'm a bit worried here. What are the pros and cons of adopting Java 11 as the minimum version?

regadas:
Hi @johnynek!
> Wait so this means all users going forward not on 2.11 have to be on Java 11 right?

that's not the intent. We want to target JVM 8 (for now) regardless of the JVM version used to buid/publish the jar. The `2.11` check there exists because that `scalac` does not support the `-release` flag only the `-target` one.

I double checked and class files for the different scala versions have `major version: 52`

regadas:
btw, this was required because the underlying plugin that we used for CI switched from 8 to 11 as default and we use I to publish the jars.

johnynek:
sorry for the latency on reviewing this. It looks right to me.

It would be interesting to think about adding some assertion that the built jars have the right class number, but I don't think we should block this (maybe a feature request for sbt-spiewak).

johnynek (approved):

Issue #329, opened 2021-12-15
A parser string interpolator


johnynek:
The idea would be implementing a string interpolator like:

```scala
val foo: Parser[A] = ...
val bar: Parser[B] = ...
val quux: Parser[C] = ...

val arrow: Parser[(A, B, C)] = parser"$foo,$bar -> $quux"
  // same as:
  // (foo <* Parser.char(','), bar <* Parser.string(" -> "), quux).tupled
```

this was suggested by @s5bug on discord

https://discord.com/channels/632277896739946517/867087707536097350/917252934524370944

PR #339, opened 2021-12-30
String interpolator parsers

stephenjudkins:
Fixes https://github.com/typelevel/cats-parse/issues/329.

I just dragged this over the finish line and am open to any suggested changes. Was a useful if somewhat painful way to dip my toes into Scala 3

### DIFF
--- a/core/shared/src/main/scala-2/cats/parse/StringInterpolation.scala
+++ b/core/shared/src/main/scala-2/cats/parse/StringInterpolation.scala
@@ -36,6 +36,7 @@ class StringInterpolationMacros(val c: whitebox.Context) {
 
   private def lit(s: String) = q"_root_.cats.parse.Parser.string($s)"
 
+
   final def parser(args: c.Expr[Any]*) =
     c.prefix.tree match {
       case Apply(_, Apply(_, parts) :: Nil) =>
@@ -44,18 +45,23 @@ class StringInterpolationMacros(val c: whitebox.Context) {
         val (matched, last) = stringParts.splitAt(args.length)
 
         val parsers: List[Tree] = matched.zip(args).map { case (a, b) =>
-          if (a.isEmpty) b.tree else q"${lit(a)} *> $b "
-        } ++ last.filter(_.nonEmpty).map(lit)
+          if (a.isEmpty) b.tree else q"${lit(a)} *> $b"
+        }
 
-        parsers match {
-          case Nil =>
-            throw new IllegalArgumentException("a non-empty string is required to create a Parser")
-          case x :: Nil => x
-          case xs =>
-            Apply(
+        val trailing = last.filter(_.nonEmpty).headOption.map(lit)
+
+        (parsers, trailing) match {
+          case (Nil, None) => throw new IllegalArgumentException("a non-empty string is required to create a Parser")
+          case (Nil, Some(t)) => t
+          case (x :: Nil, Some(t)) => q"$x <* $t"
+          case (x :: Nil, None) => x
+          case (xs, t) =>
+            val tupled = Apply(
               Select(q"_root_.cats.Semigroupal", TermName(s"tuple${parsers.length}")),
               xs
             )
+
+            t.map(p => q"$tupled <* $p").getOrElse(tupled)
         }
 
       case _ => c.abort(c.enclosingPosition, "Must be called as string interpolator")
--- a/core/shared/src/main/scala-2/cats/parse/StringInterpolation.scala
+++ b/core/shared/src/main/scala-2/cats/parse/StringInterpolation.scala
@@ -26,7 +26,7 @@ import scala.reflect.macros.whitebox
 object StringInterpolation {
 
   implicit class Helper(val sc: StringContext) extends AnyVal {
-    def parser(args: Any*): Parser[Any] = macro StringInterpolationMacros.parser
+    def parser(args: Any*): Parser0[Any] = macro StringInterpolationMacros.parser
   }
 
 }
@@ -50,8 +50,7 @@ class StringInterpolationMacros(val c: whitebox.Context) {
         val trailing = last.filter(_.nonEmpty).headOption.map(lit)
 
         (parsers, trailing) match {
-          case (Nil, None) =>
-            throw new IllegalArgumentException("a non-empty string is required to create a Parser")
+          case (Nil, None) => q"_root_.cats.parse.Parser.unit"
           case (Nil, Some(t)) => t
           case (x :: Nil, Some(t)) => q"$x <* $t"
           case (x :: Nil, None) => x
--- a/core/shared/src/main/scala-2/cats/parse/StringInterpolation.scala
+++ b/core/shared/src/main/scala-2/cats/parse/StringInterpolation.scala
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) 2021 Typelevel
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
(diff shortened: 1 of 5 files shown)


johnynek:
cc @s5bug

core/shared/src/main/scala-2/cats/parse/StringInterpolation.scala
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+package cats.parse
+
+import scala.reflect.macros.whitebox
+
+object StringInterpolation {
+
+  implicit class Helper(val sc: StringContext) extends AnyVal {
+    def parser(args: Any*): Parser[Any] = macro StringInterpolationMacros.parser
```
johnynek, line 29: I am not very familiar with macros honestly... I always have to look up details.

So, I guess that since this is a white box macro that the type gets refined (as checked in the test below)?

core/shared/src/main/scala-2/cats/parse/StringInterpolation.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  final def parser(args: c.Expr[Any]*) =
+    c.prefix.tree match {
+      case Apply(_, Apply(_, parts) :: Nil) =>
+        val stringParts = parts.collect { case Literal(Constant(s: String)) => s }
+
+        val (matched, last) = stringParts.splitAt(args.length)
+
+        val parsers: List[Tree] = matched.zip(args).map { case (a, b) =>
+          if (a.isEmpty) b.tree else q"${lit(a)} *> $b "
```
johnynek, line 47: I'm wondering about what happens if `b` is a Parser0. It seems we could often handle it (as long as there is at least one string value, or another Parser. But I guess we can't get the type of `b` right?

core/shared/src/test/scala/cats/parse/StringInterpolatorTest.scala
```diff
@@ -0,0 +1,40 @@
...
+    val foo: Parser[Boolean] = Parser.string("spam").as(true) | Parser.string("eggs").as(false)
+    val bar: Parser[BigInt] = Numbers.bigInt
+    val quux: Parser[Int] = Parser.string("quux").as(42)
+
+    val arrow: Parser[(Boolean, BigInt, Int)] = parser"$foo,$bar -> $quux"
+
+    assertEquals(
+      arrow.parseAll("spam,1234 -> quux"),
+      Right((true, BigInt(1234), 42))
+    )
```
johnynek, line 38: could we add a few assertions on `compileErrors(...)`

https://scalameta.org/munit/docs/assertions.html

it would be good to see some failures and make sure they fail in both scala 2 and 3 (although the messages may be different).

core/shared/src/test/scala/cats/parse/StringInterpolatorTest.scala
```diff
@@ -0,0 +1,40 @@
...
+
+import StringInterpolation._
+
+class StringInterpolatorTest extends munit.ScalaCheckSuite {
+
+  test("parser string interpolator") {
+    val foo: Parser[Boolean] = Parser.string("spam").as(true) | Parser.string("eggs").as(false)
+    val bar: Parser[BigInt] = Numbers.bigInt
+    val quux: Parser[Int] = Parser.string("quux").as(42)
+
```
johnynek, line 32: can we exercise with 0 args, 1 arg up to say 3 (to exercise non-tuples)
so, I guess `val x: Parser[Unit] = parser"foo"` should work. `val x: Parser[A] = parser"foo$aParser"` and then at 2 we should have tuples?

or do we require 2 or more (or 1 if we wrap with tuple1)?

johnynek (commented):
This is really awesome! I'm so excited you tackled this.

core/shared/src/main/scala-2/cats/parse/StringInterpolation.scala
johnynek: should we test that? I guess we could make sure it works as long as there is a leading `Parser` or string?

core/shared/src/test/scala/cats/parse/StringInterpolatorTest.scala
```diff
@@ -0,0 +1,40 @@
...
+package cats.parse
+
+import StringInterpolation._
+
+class StringInterpolatorTest extends munit.ScalaCheckSuite {
+
+  test("parser string interpolator") {
+    val foo: Parser[Boolean] = Parser.string("spam").as(true) | Parser.string("eggs").as(false)
+    val bar: Parser[BigInt] = Numbers.bigInt
+    val quux: Parser[Int] = Parser.string("quux").as(42)
```
johnynek, line 31: could we write this as `parser"quux".as(42)`? does parser bind tighter than method application?

s5bug:
This is awesome! Thank you so much!

core/shared/src/test/scala/cats/parse/StringInterpolatorTest.scala
stephenjudkins: Yes, it does

core/shared/src/test/scala/cats/parse/StringInterpolatorTest.scala
stephenjudkins: Wrote test to exercise these cases. Here's how it currently works:

* empty string: compile error
* non-empty string with no interpolated parsers: `Parser[Unit]` equivalent to `Parser.string("...")`
* string with single interpolated `Parser[A]`: `Parser[A]`
* string with n interpolated `Parser[{A, B, C...}]`: `Parser[(A, B, C...)]`

stephenjudkins:
Added more code to deal with `Parser0`. Turns out, in Scala 2.x since we're dealing with untyped trees things just work out like we'd expect. Scala 3 needs more type plumbing to get the same behavior working

A change from previous discussions is that `parser""` now generates a `Parser.unit:Parser0[Unit]`

core/shared/src/main/scala-2/cats/parse/StringInterpolation.scala
regadas: Yup correct, this needs to be a whitebox macro to get the type refined at each call site.

core/shared/src/main/scala-2/cats/parse/StringInterpolation.scala
```diff
@@ -0,0 +1,68 @@
...
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package cats.parse
+
+import scala.reflect.macros.whitebox
+
+object StringInterpolation {
+
+  implicit class Helper(val sc: StringContext) extends AnyVal {
```
regadas, line 28:
```suggestion
  implicit class Helper(private val sc: StringContext) extends AnyVal {
```

core/shared/src/main/scala-3/cats/parse/StringInterpolation.scala
```diff
@@ -0,0 +1,132 @@
...
+
+import scala.quoted._
+
+object StringInterpolation {
+
+  extension (inline sc: StringContext) {
+    transparent inline final def parser(inline args: Any*): Any = ${ parserImpl('sc, 'args) }
+  }
+
+  def parserImpl(sc: Expr[StringContext], args: Expr[Seq[Any]])(using q: Quotes) = {
```
regadas, line 34:
```suggestion
  private def parserImpl(sc: Expr[StringContext], args: Expr[Seq[Any]])(using q: Quotes) = {
```

core/shared/src/test/scala/cats/parse/StringInterpolatorTest.scala
```diff
@@ -0,0 +1,102 @@
...
+    val arrow: Parser[(Boolean, BigInt, Int)] = parser"$foo,$bar -> $quux"
+
+    assertEquals(
+      arrow.parseAll("spam,1234 -> quux"),
+      Right((true, BigInt(1234), 42))
+    )
+
+  }
+
+  test("including a Parser0") {
```
regadas, line 69: could we also assert  `val p: Parser0[Unit] = parser""` ?

core/shared/src/main/scala-2/cats/parse/StringInterpolation.scala
```diff
@@ -0,0 +1,68 @@
...
+    def parser(args: Any*): Parser0[Any] = macro StringInterpolationMacros.parser
+  }
+
+}
+
+class StringInterpolationMacros(val c: whitebox.Context) {
```
regadas, line 34: I believe we could make this one `private`
```suggestion
private class StringInterpolationMacros(val c: whitebox.Context) {
```

regadas (commented):
this looks great! thanks for taking this 😄 just left some minor things.

I was thinking ... would it be worth separating this into it's own module?

johnynek:
Rather than nit on this, I'm going to merge and then will iterate.

I'll send a followup PR with @regadas 's suggestions.

Thanks so much!

I look forward to a release with this!

stephenjudkins:
Well, it's functional now in but I thought on it, and: the `Parser` vs `Parser0` aspect of this is currently not ideal and inconsistent between Scala 2/3. I'm send a followup PR with some failing/non-compiling tests as a starting point for discussion

johnynek:
@stephenjudkins take a look at my PR where I also tried to factor the code similarly and added a few more tests.

PR #340, opened 2021-12-31
Return Parser if parser macro finds a non-empty string

johnynek:
This adds some tests of return types and tries a bit harder to get a `Parser` back.

### DIFF
--- a/core/shared/src/main/scala-2/cats/parse/StringInterpolation.scala
+++ b/core/shared/src/main/scala-2/cats/parse/StringInterpolation.scala
@@ -43,6 +43,8 @@ private class StringInterpolationMacros(val c: whitebox.Context) {
         val stringParts = parts.collect { case Literal(Constant(s: String)) => s }
         val stringHead = stringParts.head
 
+        // If there is at least one Parser, we can return a Parser, else Parser0
+        val returnParser = stringParts.exists(_.nonEmpty)
         val unitParser = q"_root_.cats.parse.Parser.unit"
         def all(xs: List[Tree]): Tree =
           xs match {
@@ -87,8 +89,9 @@ private class StringInterpolationMacros(val c: whitebox.Context) {
             stringAfter(suffix, last)
           } else {
             // both the head and tail are empty strings
-            // TODO: if there are any internal strings we could also return a Parser
-            suffix
+            // this is uglier, but should still be valid
+            if (returnParser) q"$suffix.unsafeCastToParser"
+            else suffix
           }
         }
       case _ => c.abort(c.enclosingPosition, "Must be called as string interpolator")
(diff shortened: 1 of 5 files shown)


johnynek:
cc @stephenjudkins @s5bug @regadas

johnynek:
follow up to #339

stephenjudkins:
This is definitely an improvement! But it doesn't solve all the cases and I feel we're heading deeper into more corners and a there's a simpler, more general strategy that would work. The `def tupleOf(lst: List[TypeTree]): TypeTree` seems especially gnarly...

johnynek:
@stephenjudkins yeah, the `tupleOf` seems like there should be a standard way to build that, but I couldn't find it in a bit of looking.

johnynek:
Here's a wild idea: parse the strings as regular expressions.

So you could do things like `parser"$foo ->.*"`

If the regular expressions supported were a strict subset of valid Java regular expressions it might not be too confusing.

You could even do non-nested capture using the `.string` method on parsers.

s5bug:
You could support the subset of RegEx that is supported by [antimirov](https://github.com/non/antimirov), perhaps?

I'm worried about the likelihood of `parser"$foo($bar)"` and `parser"$foo/$bar"`. And what would you even do with `[$foo]`?

stephenjudkins:
Well, I had something that looked pretty nice and elegant in Scala 3 until I needed to construct an anonymous function, at which point I hit a brick wall

stephenjudkins:
FYI, I got further down the Scala 3 path and found a strategy that should get what we want. I'm feeling a bit better about Scala 3 macros, and the fact we can see the types of what we're passing in actually enables a strategy that'd be impossible in Scala 2

johnynek:
@stephenjudkins should we merge this and iterate from there, or what do you think?

One thing I think we should consider slightly. I definitely want the two to not conflict on a big set of cases (otherwise we can't easily write cross compiling code), but I can imagine allowing some to compile on scala 3 that don't compile on scala 2 (which we can't figure out how to do) for users only on scala 3.

regadas (approved):
This looks good to me. Took me some time to get around it 😅 

This got me thinking about cost benefits. Somehow I fear that the fancier we get with this the harder will be to maintain it and keep feature parity across scala versions. idk it might just be me overthinking it and this is totally fine.

johnynek:
I'm kind of on the fence on this one the more I think...

Can we post some compelling examples here? The problem is in many (most?) parsers you actually want to be generous on accepting whitespace, so parsing `a -> b` you would want `parser"$p${ws}->${ws}$p"` or something, which maybe not really better than `(p <* ws, string("->") *> ws *> p).tupled` or something.

A second thing I'm nervous about: I find `.soft ~` to be almost as common a strict/hard product but you often think about each one. The parser macro is doing a hard product, but I fear that will lead people to `.backtrack` the whole thing working against a goal of this library: to guide you towards writing safe and high performance parsers that point directly at your error.

I would welcome thoughts on this before we publish a version with this and feel stuck with the API.

PR #341, opened 2021-12-31
Tests to clarify Parser/Parser0 distinction when using string interpolator

stephenjudkins:
These new tests identify two problems:

* Interpolation of both a `Parser0` and `Parser` without any string parts should still expand to a `Parser`. However, using `tupleN` to combine our parsers won't work as a strategy because it finds the LUB of every component which ends up being `Parser0`. This is consistent across Scala2 and Scala 3
* Interpolation of a `Parser0` preceded by a string part should expand to a `Parser`, but doesn't in Scala 3. This is an artifact of the janky home-grown "type inference" I wrote. We should mark a sub-parser as still being a `Parser` when it has a preceding string part because it'll get turned into one when we do `{lit} *> p` anyways


johnynek:
check out #341 

I solved some of these but not all of these issues.

PR #347, opened 2022-01-08
Optimize .as(()) to .void

johnynek:
People sometimes write .as(()) and this makes sure that is exactly the same as .void.


johnynek:
same as #346 but GitHub wasn't seeing the second push to the branch

rossabaker (approved):

PR #348, opened 2022-01-09
introduce hasKnownResult to optimize several parsers.

johnynek:
Often, we can statically see what the result value will be if a parser succeeds.

If we know this, can optimize map, select and flatMap.

In particular, we can avoid reallocating a parser inside each flatMap.

Most users should avoid using flatMap anyway, but in a library function, you might need flatMap but want to get the optimization when the user passes you a constant parser.


regadas (approved):
It will be interesting to see new benchmark numbers.

regadas:
@johnynek could you rebase this?

johnynek:
Yes! I will do. Thanks for the review.

Note, this optimization will mostly apply to flatMap, and most parser don't need flatMap (e.g. our json benchmarks don't use flatMap). But still, I thought it was worth trying to limit the worst case.

johnynek:
@regadas let me know if this looks good.

regadas (approved):
Looks great!

Issue #350, opened 2022-01-12
RadixNode and stringIn are a bit slow


johnynek:
We have a benchmark for stringIn, which exercises stringIn. I also compared it to a naive iteration through a list to check if we beat that, and we don't.

I think there are a few issues:

1. we use binary search to find if a character is in a list, if there are only a few possible strings, the cost of the binary search might be comparable to checking the few strings. An alternative would be to use an indexing scheme so we can compute the index into an array from the char value.
2. stringIn knows the full strings that can match, but it discards them, so it has to allocate at the end again. If we built the tree storing the fully matching value so we don't need to reallocate a string copy we will get a significant win. When I benchmark `stringIn(s).void` in our current benchmark it almost matches the naive loop above (it should beat it), but without the void it is about 2x slower.

approach should be to move all the code to matching the RadixNode into that file (we are splitting the code now and some of it lives in Parser). Then we should scalatest very completely that matching code. Finally we should add a benchmark that compares `stringIn` when there are a much longer list of strings over a longer input that will match. The current benchmark is pretty hostile since the naive thing is very close to optimal.

Finally, with that in place, we should experiment with the two suggestions above to see if they can actually solve the problem.

PR #352, opened 2022-01-16
Improve RadixNode add a few more benchmarks

johnynek:
part of #350 

This may be marginally faster, but the main value it adds is that RadixNode can now return the matching string without allocating. Also, I think the implementation is somewhat simpler.

I expect this should be faster if the set of strings gets larger (the benchmarks only have a few different prefixes. This uses a hash based approach so it is always O(1) to check a character (not log(N) worst case).

```
Before:

[info] Benchmark                         Mode  Cnt   Score   Error  Units
[info] StringInBenchmarks.linearMatchIn  avgt    3  77.603 ± 0.529  ns/op
[info] StringInBenchmarks.radixMatchIn   avgt    3  59.675 ± 2.126  ns/op

This PR:

[info] Benchmark                         Mode  Cnt   Score   Error  Units
[info] StringInBenchmarks.linearMatchIn  avgt    3  77.427 ± 4.083  ns/op
[info] StringInBenchmarks.radixMatchIn   avgt    3  59.067 ± 4.797  ns/op
```

### DIFF
--- a/core/shared/src/main/scala/cats/parse/RadixNode.scala
+++ b/core/shared/src/main/scala/cats/parse/RadixNode.scala
@@ -25,8 +25,6 @@ import cats.data.NonEmptyList
 import cats.kernel.Semilattice
 import scala.annotation.tailrec
 
-import cats.syntax.all._
-
 private[parse] final class RadixNode(
     matched: String,
     bitMask: Int,
@@ -114,10 +112,10 @@ private[parse] object RadixNode {
       val branching = bitMask + 1
       val prefixes = new Array[String](branching)
       val children = new Array[RadixNode](branching)
-      val tree = nonEmpties.groupByNel { s => (s.head.toInt & bitMask) }
+      val tree = nonEmpties.groupBy { s => (s.head.toInt & bitMask) }
       tree.foreach { case (idx, strings) =>
         // strings is a NonEmptyList[String] which all start with the same char
-        val prefix1 = strings.reduce(commonPrefixSemilattice)
+        val prefix1 = strings.reduce(commonPrefixSemilattice.combine(_, _))
         val plen = prefix1.length
         val node = fromTree(thisPrefix, prefix + prefix1, strings.map(_.drop(plen)).toList)
         prefixes(idx) = prefix1
@@ -129,7 +127,7 @@ private[parse] object RadixNode {
   }
 
   def fromSortedStrings(strings: NonEmptyList[String]): RadixNode =
-    fromTree(null, "", strings.distinct.toList)
+    fromTree(null, "", strings.toList.distinct)
 
   def fromStrings(strs: Iterable[String]): RadixNode =
     fromTree(null, "", strs.toList.distinct)
--- a/core/shared/src/main/scala/cats/parse/RadixNode.scala
+++ b/core/shared/src/main/scala/cats/parse/RadixNode.scala
@@ -39,6 +39,22 @@ private[parse] final class RadixNode(
     s"RadixNode($matched, $bitMask, $ps, $cs)"
   }
 
+  /** @return
+    *   all strings that are in this RadixNode
+    */
+  def allStrings: List[String] = {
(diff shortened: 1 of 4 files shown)


johnynek:
I added a benchmark that is close to ideal for the radix tree to have a kind of upper bound comparison. In that example this PR gives performance about 27% faster than main (and 90x faster than a naive linear search).

So, I think it is fair to say the worst case performance of the new code is slightly worse than a linear search (17% comparing the favorable linear search benchmark with the tree). Note, linear search can be very good with only a few alternatives. But when there are many alternatives, the tree can be far, far better.
```
// With new benchmarks:
[info] Benchmark                         (test)  Mode  Cnt      Score      Error  Units
[info] StringInBenchmarks.linearMatchIn     foo  avgt    3     77.718 ±    3.173  ns/op
[info] StringInBenchmarks.linearMatchIn   broad  avgt    3  92973.234 ± 2156.151  ns/op
[info] StringInBenchmarks.oneOfParse        foo  avgt    3     91.240 ±    4.854  ns/op
[info] StringInBenchmarks.oneOfParse      broad  avgt    3    806.046 ±   37.539  ns/op
[info] StringInBenchmarks.radixMatchIn      foo  avgt    3     59.750 ±    1.315  ns/op
[info] StringInBenchmarks.radixMatchIn    broad  avgt    3    769.000 ±   25.609  ns/op
[info] StringInBenchmarks.stringInParse     foo  avgt    3     91.551 ±    0.600  ns/op
[info] StringInBenchmarks.stringInParse   broad  avgt    3    809.561 ±   54.042  ns/op

// new benchmarks on main
[info] Benchmark                         (test)  Mode  Cnt      Score      Error  Units
[info] StringInBenchmarks.linearMatchIn     foo  avgt    3     78.476 ±    2.503  ns/op
[info] StringInBenchmarks.linearMatchIn   broad  avgt    3  93335.231 ± 2911.905  ns/op
[info] StringInBenchmarks.oneOfParse        foo  avgt    3     96.497 ±    1.285  ns/op
[info] StringInBenchmarks.oneOfParse      broad  avgt    3   1107.582 ±    1.940  ns/op
[info] StringInBenchmarks.stringInParse     foo  avgt    3     96.489 ±    0.397  ns/op
[info] StringInBenchmarks.stringInParse   broad  avgt    3   1106.763 ±    3.840  ns/op
```

johnynek:
@regadas if you have time, I'd love your review (or anyone).

I think this is ready.

regadas:
Hi, @johnynek this one slipped under my radar. I'll take a look at it tmr.

johnynek:
Thank you! I know this OSS stuff can be a grind. Don't worry at all. I appreciate your help.

regadas (approved):
This looks great! Sorry for the delay @johnynek

core/shared/src/main/scala/cats/parse/RadixNode.scala
```diff
@@ -22,85 +22,170 @@
...
+    *   the string to match against this RadixNode
+    * @param offset
+    *   the initial offset
+    * @return
+    *   the new offset after a match, or -1
+    */
+  def matchAt(str: String, off: Int): Int =
+    matchAtOrNull(str, off) match {
+      case null => -1
+      case nonNull => nonNull.length
```
johnynek, line 70: Actually I think this is wrong. It has to be offset + NonNull.length, or the comment is wrong. Maybe we should just fix the comment, but also the Parser is assuming it returns the new offset not the length.

I'll send a PR with a failing test and then fix.

I think this is happening because we are only testing at offset 0.

regadas: Nice catch!

Being `offset + nonNull.length` makes more sense and the consistency with `matchAtOrNull` reflects it (which needs a fix).

These make more sense now 
```scala
            assert(len <= targ.length, s"len = $len, off = $off")
            assertEquals(left, targ.substring(off, len), s"len = $len, left = $left")
```

PR #354, opened 2022-01-20
Migrate to sbt-typelevel

armanbilge:
Migrates the build to sbt-typelevel. It's very similar to sbt-spiewak, so most things are the same. Notable differences:
- Snapshots are now published to Sonatype snapshots instead of Maven. Unless you like them at Maven, in which case easy to restore.
- Java 8 is the default JDK for CI (the build was using JDK 11, was this intentional?).
- MiMa is now enabled for Scala 3 and Scala.js.
- It is now safe to tag a release immediately after merging a PR/pushing to main without causing a publication failure.

Since the build here is fairly customized, I've been conservative about adopting additional sbt-typelevel features (e.g., CI matrices crossed on JVM/JS in addition to Scala version).


build.sbt
```diff
@@ -5,21 +5,14 @@ import Dependencies._
 addCommandAlias("fmt", "; scalafmtAll; scalafmtSbt")
 addCommandAlias("fmtCheck", "; scalafmtCheckAll; scalafmtSbtCheck")
 
-addCommandAlias("prePR", "; githubWorkflowGenerate ; +fmt; bench/compile; +test")
+tlReplaceCommandAlias("prePR", "; githubWorkflowGenerate ; +fmt; bench/compile; +test")
```
armanbilge, line 8: sbt-typelevel defines its own `prePR` so we replace the alias.

build.sbt
```diff
@@ -152,19 +133,19 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
...
     crossScalaVersions := (ThisBuild / crossScalaVersions).value.filterNot(_.startsWith("2.11")),
-    Global / scalaJSStage := FastOptStage,
-    parallelExecution := false,
-    jsEnv := new org.scalajs.jsenv.nodejs.NodeJSEnv(),
-    // batch mode decreases the amount of memory needed to compile scala.js code
-    scalaJSLinkerConfig := scalaJSLinkerConfig.value
-      .withBatchMode(scala.sys.env.get("TRAVIS").isDefined)
-      .withModuleKind(ModuleKind.CommonJSModule),
-    coverageEnabled := false,
-    scalaJSUseMainModuleInitializer := false
```
armanbilge, line 167: Pretty sure these are all redundant.

build.sbt
```diff
@@ -66,19 +59,8 @@ ThisBuild / githubWorkflowPublish ++= Seq(
...
 )
 
-ThisBuild / homepage := Some(url("https://github.com/typelevel/cats-parse"))
-
-ThisBuild / scmInfo := Some(
-  ScmInfo(
-    url("https://github.com/typelevel/cats-parse"),
-    "git@github.com:typelevel/cats-parse.git"
-  )
-)
```
armanbilge, line 76: sbt-typelevel sets this automatically.

armanbilge (commented):
Ready for review. Thanks!

johnynek (approved):
Thanks for doing this!

Cc @mpilquist who set up the build.

PR #359, opened 2022-01-27
Use scalafmt scala213 runner dialect

regadas:
It's ok to use it and for metals users, it will avoid the nagging.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,6 +1,6 @@
 version = "3.3.3"
 
-runner.dialect = scala211
+runner.dialect = scala213
 
 style = default
 


regadas (commented):
Ok I was puzzled with why would this fail 😄 but it seems that mima on master is also unhappy ... I'll take a look at this later today

armanbilge:
Ah, this is because until https://github.com/typelevel/cats-parse/pull/354 you were not checking MiMa on Scala 3 and https://github.com/typelevel/cats-parse/pull/352 hadn't merged in the latest changes on main.

In my experience most MiMa failures that only appear on Scala 3 are harmless and only because of https://github.com/lightbend/mima/issues/682.

johnynek:
Thanks for chiming in @armanbilge 

@regadas I can send a PR to filter the changes. Then you can merge master. You shouldn't have to fix my issues.

PR #360, opened 2022-01-27
Fast charIn for ranges like 'A' to 'Z'

Odomontois:


johnynek (approved):
This is great! Thank you!

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -1533,16 +1534,22 @@ object Parser {
...
+      case range: NumericRange.Inclusive[Char] if range.step.toInt == 1 =>
+        val start = range.start.toInt
+        val end = range.end.toInt
+        val bitSet = BitSetUtil.bitSetForRange(end - start + 1)
+        Impl.CharIn(start, bitSet, NonEmptyList.one(range.start -> range.end))
+      case _ =>
+        val ary = cs.toArray
+        Arrays.sort(ary)
+        rangesFor(ary) match {
+          case NonEmptyList((low, high), Nil) if low == Char.MinValue && high == Char.MaxValue =>
```
johnynek, line 1548: actually, we aren't doing this optimization above are we? if `range.start == 0` and `range.end == 0xffff` we should return `anyChar` right?

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1181,6 +1181,18 @@ class ParserTest extends munit.ScalaCheckSuite {
...
 
+  property("charIn range matches charIn list") {
+    forAll { (c1: Char, c2: Char, str: String) =>
+      val start = c1.min(c2)
+      val end = c2.max(c2)
+
+      val p1 = Parser.charIn(start to end)
+      val p2 = Parser.charIn((start to end).toList)
+
+      assertEquals(p1.parse(str), p2.parse(str))
```
johnynek, line 1192: actually, I think we want it stronger: `assertEquals(p1, p2)` and not bother taking the `str` as input.

core/shared/src/test/scala/cats/parse/ParserTest.scala
```diff
@@ -1181,6 +1181,18 @@ class ParserTest extends munit.ScalaCheckSuite {
     }
   }
 
+  property("charIn range matches charIn list") {
+    forAll { (c1: Char, c2: Char, str: String) =>
+      val start = c1.min(c2)
+      val end = c2.max(c2)
```
johnynek, line 1187: I think you want `c1.max(c2)`. This is always returning `c2`.

johnynek (changes requested):
left some comments.

The tests failed but I didn't quite see the error yet, but I think there likely is one.

Also, master got a fix for a CI check that would have also likely failed for you. Please merge master if you can.

Thanks again for the PR!

core/jvm/src/main/scala/cats/parse/BitSet.scala
```diff
@@ -64,7 +64,7 @@ object BitSetUtil {
 
   def bitSetForRange(count: Int): BitSet = {
     val bs = new BitSet(count)
-    bs.flip(0, count - 1)
```
Odomontois: That was an actual source of the failed test. BitSet.flip upper bound seems to be exclusive despite being named  `toIndex`

johnynek (approved):
Thank you!

PR #373, opened 2022-02-11
Add fromCharMap and fromStringMap

johnynek:
These are pretty useful and save the user from using `apply` on a Map which may appear unsafe to a reviewer in user code, but is safe in this context (we know we return a matching key).


regadas (approved):
Looks great and useful!

PR #376, opened 2022-02-13
wip:add native build

i10416:
close #371 

Waiting for non-snapshot cats-core_native0.4_3 published ~which depends on~
- ~https://github.com/typelevel/scalacheck/pull/868~
and test dependencies for native 3.x, which depends on
- ~https://github.com/typelevel/scalacheck/pull/868~
- ~https://github.com/scalameta/munit/pull/477~
~and they are waiting for the compiler flag to be introduced in Scala 3.1.2.~

Now, cats(snapshot version) supports Scala Native x Scala 3
- https://github.com/typelevel/cats/pull/4228

PR #384, opened 2022-02-18
update: use java.util.BitSet on js platform

### LINKED ISSUE #381: use java.util.BitSet on scalajs
1.9.0 supports java.util.BitSet now, so we could remove the different code between jvm and js.

https://www.scala-js.org/news/2022/02/14/announcing-scalajs-1.9.0/

i10416:
close #381 
scala.js supports java.util.BitSet since 1.9.0


johnynek:
I don't think we can merge this and #376.

Also the tests will be very slow if we don't have the switch for isScalaJs.

Can scala-native handle Java util Bitset?

Finally, this will be binary incompatible.

johnynek:
But *thanks* for the PR!!

Hopefully we can merge some version.

i10416:
Thank you for review. I will tweak what you pointed out.

> I don't think we can merge this and #376

Yes, I think too.
I planed to share source between jvm/js first, then tweak #376 

> Can scala-native handle Java util Bitset?
 
AFAIK, scala native does not support java.util.BitSet :(

i10416:
> Also the tests will be very slow if we don't have the switch for isScalaJs.

I think it would be better to switch scala.js/jvm tests based on build settings (e.g. generate util object as [sbt-buildinfo](https://github.com/sbt/sbt-buildinfo) do ) rather than `BitSetUtil` fields. What do you think?

i10416:
Ah, I found https://github.com/portable-scala/sbt-crossproject/issues/17. IIUC, we need to copy-paste `BitSetUtil` between jvm and js to add native platform impl later?

armanbilge:
@i10416 I'm actually about to submit a PR to fix https://github.com/portable-scala/sbt-crossproject/issues/17 :)

> AFAIK, scala native does not support java.util.BitSet :(

That's true, but now that Scala.js does you can probably just copy-paste the implementation and tests. That's how many things get ported to native :)

> I think it would be better to switch scala.js/jvm tests based on build settings

I think if you have a dependency on cats-laws in the tests you can just use this:

https://github.com/typelevel/cats/blob/b307688c391a47dc6dd82d42c15edf273bf2471b/kernel-laws/jvm/src/main/scala/cats/platform/Platform.scala

armanbilge:
- https://github.com/scala-native/scala-native/pull/2565

i10416:
> tests will be very slow if we don't have the switch for isScalaJs.

Now that test iteration is switched based on js/jvm, what to do are

1. merge #384 in future release(or is there a nice trick to keep bincompat?)
2. wait for https://github.com/scala-native/scala-native/pull/2565 to be merged/released(thx @armanbilge )
3. then tweak and finish #376.

johnynek:
I think it is reasonable to release `0.4.0` and break this minor binary compatibility (isScalaJs, isScalaJvm)...

Those really should have been made `private [cats.parse]` anyway...

armanbilge:
FWIW I'd appreciate not bumping and preserving binary-compatibility here instead. It makes downstream cross-builds slightly more complicated when the different platforms are pinned on different versions of the same dependency.

Apologies if I am missing something, but why is it so hard to preserve bincompat here? Or is it just undesirable?

johnynek:
@armanbilge I'm confused by what you mean.

The only real break to binary compatibility is the symbols `isScalaJvm` and `isScalaJs`. I bet no one is actually using those. So, any real dependency won't be broken.

The other errors are for items inside the `private object Impl` which no one can see without doing some weird reflection hacks, which we don't promise to support.

Do you think there are downstream users that care about the first two?

Or is it something about using scalajs 1.9.0?

armanbilge:
FTR I completely agree that its unlikely that removing those will break any downstreams. 

However, it's the version bump that got my attention: it will force downstreams to also bump their versions to take on this update.

johnynek:
okay, fair point.

Let's add the isScalaJvm, isScalaJs back even though it is inelegant. We can add exclusions for the changes inside `private object Impl`.

i10416:
Thank you for reviewing. I will put isScalaJvm and isScalaJs back.

core/shared/src/main/scala/cats/parse/BitSet.scala
```diff
@@ -22,12 +22,12 @@
 package cats.parse
 
 import java.util.BitSet
-
 object BitSetUtil {
   type Tpe = BitSet
-
+  @deprecated("", "0.3.7")
```
i10416, line 27: I changed  here to `isScalaJVM =false , isScalaJs = false` because I think combination of `isScalaJVM =true, isScalaJs = false`, `isScalaJVM =false, isScalaJs = true` or `isScalaJVM =true, isScalaJs = true` does not make sense.

i10416:
Hmmm, can't I write exclude rules like `cats.parse.Parser#Impl.*` ?

core/shared/src/main/scala/cats/parse/BitSet.scala
i10416: > instead we could move the BitSet code to ...

Yeah, actually, I thought my idea was not nice too😓

That seems better solution! Thanks!

core/shared/src/test/scala/cats/parse/LocationMapTest.scala
```diff
@@ -26,7 +26,7 @@ import Prop.forAll
 
 class LocationMapTest extends munit.ScalaCheckSuite {
 
-  val tests: Int = if (BitSetUtil.isScalaJs) 50 else 20000
+  val tests: Int = 20000
```
i10416, line 29: Oh, I forgot replacing here with TestUtils.iteration😓

i10416:
I removed deprecated annotations, which I added just in case though I didn't think someone uses `isScalaJs` and `isScalaJvm` api, because I assume you don't like warning on the ground that you regard warning as fatal when checking source before marge (while I felt it makes sense to show deprecated warning).

i10416:
There remain some mima warnings such as `method bitSetFor(Array[Char])scala.collection.mutable.BitSet in object cats.parse.BitSetUtil has a different result type in current version, where it is java.util.BitSet rather than scala.collection.mutable.BitSet filter with: ProblemFilters.exclude[IncompatibleResultTypeProblem]("cats.parse.BitSetUtil.bitSetFor")`. 

Should I suppress them too for now?  If we would suppress them, I think we would need to remove the exclude rules in a future release as this change is from 0.3.6 to 0.3.7 only.

johnynek:
yeah, we can suppress those errors.

They are only for scalajs and they are only if for some weird reason users accessed that BitSet abstraction code, and I don't know why they would. So, I would bet it will impact exactly zero users.

Also, let's add an issue to remove isScalaJs/isScalaJvm when we release a breaking change.

core/shared/src/test/scala/cats/parse/BitSetTest.scala
```diff
@@ -24,13 +24,6 @@ package cats.parse
 import org.scalacheck.Prop.forAll
 
 class BitSetTest extends munit.ScalaCheckSuite {
-  test("isScalaJs/isScalaJvm is consistent") {
```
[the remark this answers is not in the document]
i10416: sure

i10416:
@johnynek 
Can I ask you to review this PR? 

Just a reminder. I'm not in a hurry, of course :)

johnynek:
Sorry I missed that you had updated this. It looks good to me. If CI is green I'll merge. Thanks so much for working so much to get this done within the binary compatibility constraints.

johnynek (approved):

i10416:
Sooooooorry! I will fix this by tommow! 🙇

johnynek:
huh... the CI does not seem to be running.

i10416:
Strang...

johnynek:
Thanks for pushing this one through!

Issue #382, opened 2022-02-17
`p1.? ~ p2`  !=  `(p1 ~ p2) | p2` for error reporting purposes


carueda:
Hello there!

Assume these base parsers:

```scala
val p1 = P.char('a')
val p2 = P.char('z')
```

Consider now these:

```scala
val p = p1.? ~ p2
val q = (p1 ~ p2) | p2
```

(that is, both accept the same input space: `{ "az", "z" }`).

With an empty input string, `p` and `q` generate these errors:

```scala
p.parse("") ==> Left(
          Error(
            0,
            NonEmptyList(
              InRange(0, 'z', 'z'), 
              List()
            )
          )
        )

q.parse("") ==> Left(
          Error(
            0,
            NonEmptyList(
              InRange(0, 'a', 'a'),
              List(InRange(0, 'z', 'z'))
            )
          )
```

For error reporting purposes I prefer the latter response as it indicates that not only 'z' but also 'a' can appear at the point of failure.

So, for alternative parser cases like the above, I could just write them à la `q` to accomplish this "enriched" error reporting.

However, do I have to?  It seems unnecessary and I would prefer to just write them in the `p` style example. Am I missing something? Thanks!

johnynek:
Thanks for the issue. So, you are raising a really good point.

your first parser `p` is really the same as `(p1.map(Some(_)) | pure(None)) ~ p2`. So you want some distributive law: `(a | b) ~ c == (a ~ c) | (b ~ c)`

We have a version of that law here:

https://github.com/typelevel/cats-parse/blob/455e9791ffc75a8e726ffd1204e2eda0277b0383/core/shared/src/test/scala/cats/parse/ParserTest.scala#L1849

Which is a bit more complex, but it is only testing the results, not that the error messages are identical.

Maybe the issue in this case is because `p1.?` can always succeed: so from the parser's point of view, it successfully parsed `p1.?` by returning `None`. I can imagine something like an optimization rule such that if you have a union that ends with something that always succeeds, then push the product down:

```
def product[A, B](left: Product0[A], right: Product0[B]): Product0[(A, B)] =
  (left, right) match {
    case OneOf0(items), right if alwaysSucceeds(items.last) =>
      OneOf0(items.map { li => product(li, right) })
    case _ =>
      // do the regular thing.
  }
```
but this isn't quite right for the same reason we don't have the law `(a | b) ~ c == (a ~ c) | (b ~ c)` in general.

[...]

PR #386, opened 2022-02-21
Add the ?: operator from #382

johnynek:
adds the operator conceived by @carueda in #382 and some associated laws.

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Numbers.scala
+++ b/core/shared/src/main/scala/cats/parse/Numbers.scala
@@ -51,7 +51,7 @@ object Numbers {
   /** A nonNegativeIntString possibly preceded by '-'
     */
   val signedIntString: Parser[String] =
-    (Parser.char('-').?.with1 ~ nonNegativeIntString).string
+    (Parser.char('-') ?: nonNegativeIntString).string
 
   /** map a signedIntString into a BigInt
     */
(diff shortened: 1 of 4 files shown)


carueda:
Wonderful, thanks!!

carueda (approved):
Only comment is that there are no tests verifying expected errors, but I understand that's the general state of affairs, so no show-stopper at all.  Many thanks again for incorporating `?:` in this amazing library!

johnynek:
The more I think about this one the more I am not sure...

pros:
  1. useful for clearer error messages
  2. allows avoiding introducing Parser0 to get left optionality.

cons:
  1. possibly cryptic syntax for users `?:` which part is optional left or right (left, same side as ?).
  2. methods ending in `:` can be confusing for new scala users (they are right associative).
  3. possibly we could solve this so that `foo.? ~ bar` does the right thing. We could potentially rewrite that so that. The questions, would it be safe to rewrite a `OneOf0(ls) ~ r` into `OneOf(ls.map(_.with1 ~ r))`? I thought maybe it wouldn't before since I had a similar law that failed. Even failing that, it could be that we can rewrite a `OneOf0(...)` ending in `Pure` to a `(oneOf(ls.init) ~ r) | (ls.last ~ r)`... If that rewrite is lawful, then this method isn't needed and we could just fix underlying syntax.

I'll experiment with the weaker rewrite in 3 in some laws. I am pretty sure the strong rewrite fails due to lack of backtracking, but maybe the weak one passes since `Pure` always backtracks and it is the last item in the oneOf.

carueda:
I agree that `?:` could create unnecessary confusion for some users, especially given the other regular operators in it (in particular, the left-associativity of the common `~`). No need to incorporate the new operator.  I think users with that kind of need can just add an extension or something (as I'm doing already with just a couple of lines).

But I vote for somehow supporting enriched error reporting for `p1.? ~ p2.? ~ p3.? ~ p4.? ~ p5.? ~ pz`.

PR #389, opened 2022-02-27
Optimize `a.? ~ b` for  better error reporting

johnynek:
replacement for #386 related to #382 

@carueda

please take a look at this.

I ran these new laws (before optimization) with the count turned up very high. I *think* they are legitimate and address the use case you care about.


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -78,34 +78,19 @@ sealed abstract class Parser0[+A] { self: Product =>
...
 
+  private[this] lazy val allParser = (this ~ Parser.end)
+
   /** Attempt to parse all of the input `str` into an `A` value.
     *
     * This method will return a failure unless all of `str` is consumed during parsing.
     *
     * `p.parseAll(s)` is equivalent to `(p <* Parser.end).parse(s).map(_._2)`.
     */
-  final def parseAll(str: String): Either[Parser.Error, A] = {
```
johnynek, line 87: this old implementation was actually hiding some error messages which the optimization of `a.? ~ b` discovered. Replaced with a direct implementation.

carueda:
@johnynek 

Thanks for this!

LGTM from the user perspective, that is, given that the "check the issue #382 example" test added here is (consistently) passing on my computer.

By "consistently" I mean that this particular test always passed after a couple of `> test` runs. However, I got failing tests in one of such runs, seed = `"ZR3wyYYbvoeJ2-SQF9Kzu08i34qM2BRhCHUtjM7TImK="`, including these:

```
...
==> X cats.parse.ParserTest.a.flatMap(b) composes as expected parser01  0.141s munit.FailException: /Users/carueda/github/cats-parse/core/shared/src/test/scala/cats/parse/ParserTest.scala:1025 stringSetB = Set(i>$) stringSetA = Set()
1024:              assert(charSetB.subsetOf(charSetA), s"charSetB = $charSetB charSetA = $charSetA")
1025:              assert(
1026:                stringSetB.subsetOf(stringSetA),

Failing seed: ZR3wyYYbvoeJ2-SQF9Kzu08i34qM2BRhCHUtjM7TImK=
You can reproduce this failure by adding the following override to your suite:

  override val scalaCheckInitialSeed = "ZR3wyYYbvoeJ2-SQF9Kzu08i34qM2BRhCHUtjM7TImK="

Falsified after 129 passed tests.
> ARG_0: GenT(OneOf(List(Map(Void(AnyChar),<function1>), Map(Str(i>$),<function1>))))
> ARG_1: ""
> ARG_2: org.scalacheck.GenArities$$Lambda$12690/1146395504@2c5eb019
    at munit.FailException.copy(FailException.scala:27)
    at munit.FailException.withMessage(FailException.scala:20)
    at munit.FailException.withMessage(FailException.scala:3)
    at munit.ScalaCheckSuite.propToTry(ScalaCheckSuite.scala:91)
    at munit.ScalaCheckSuite.$anonfun$scalaCheckPropTransform$3(ScalaCheckSuite.scala:46)
    at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:475)
    at munit.Suite$$anon$1.execute(Suite.scala:26)
    at scala.concurrent.impl.Promise$Transformation.submitWithValue(Promise.scala:429)
    at scala.concurrent.impl.Promise$DefaultPromise.submitWithValue(Promise.scala:338)
    at scala.concurrent.impl.Promise$DefaultPromise.dispatchOrAddCallbacks(Promise.scala:312)
    at scala.concurrent.impl.Promise$DefaultPromise.transform(Promise.scala:128)
    at munit.internal.FutureCompat$ExtensionFuture.transformCompat(FutureCompat.scala:16)
    at munit.ScalaCheckSuite.$anonfun$scalaCheckPropTransform$2(ScalaCheckSuite.scala:48)
    at munit.GenericTest.$anonfun$withBodyMap$1(GenericTest.scala:33)
    at munit.MUnitRunner.$anonfun$runTestBody$1(MUnitRunner.scala:296)
...
==> X cats.parse.ParserTest.parseAll law  0.028s munit.ComparisonFailException: /Users/carueda/github/cats-parse/core/shared/src/test/scala/cats/parse/ParserTest.scala:1961
1960:
1961:      assertEquals(a.fa.parseAll(str), pall)
1962:    }
values are not the same
=> Obtained
Left(
  value = Error(
    failedAtOffset = 0,
    expected = NonEmptyList(
      head = OneOfStr(
        offset = 0,
        strs = List(
          "i>$"
        )
      ),
      tail = List(
        InRange(
          offset = 0,
          lower = '\u0000',
          upper = '\uffff'
        )
      )
    )
  )
)
=> Diff (- obtained, + expected)
     expected = NonEmptyList(
-      head = OneOfStr(
+      head = InRange(
         offset = 0,
-        strs = List(
-          "i>$"
-        )
+        lower = '\u0000',
+        upper = '\uffff'
       ),
-      tail = List(
-        InRange(
-          offset = 0,
-          lower = '\u0000',
-          upper = '\uffff'
-        )
-      )
+      tail = Nil
     )
```

johnynek:
@carueda that test failure was against: [beaf960](https://github.com/typelevel/cats-parse/pull/389/commits/beaf960c7a8067a736dd95244a1cd807ab7e9ecf)?

johnynek:
I have run with 20k iterations and I haven't seen that since I made this change: https://github.com/typelevel/cats-parse/pull/389#discussion_r815377008

carueda:
I added my previous comment and then got an update notification here... so I then updated my clone and, yes, I was just able to reproduce with that seed on top of the last commit:

```
commit 055bd1ceaa6861641baea4964976362dbc4401fa (HEAD -> oscar/optimize_oneof_product, origin/oscar/optimize_oneof_product)
...
    add a test for chains of options
```

johnynek:
well, this is pretty frustrating...

So, `select` and `map` trigger some optimizations if the result is known, which is the code path this is hitting, but `oneOf` does not do the same optimizations, which is the cause of these failures.

I think it is a case of the result being correct but it reporting redundant error information. So it is saying I expected `anyChar` or a specific string. But, if the `anyChar` doesn't match, the specific string can never match, so that is redundant information, but it was hidden under a `Map` in one case, which got friend by other optimizations.

johnynek:
@carueda I *think* we have it now. Care to take another look?

carueda:
Thanks again!

All tests are passing on my side, also when using that seed that exposed the regression before.  Good!

However, I note the following wrt the "series of options: pa.? ~ pb.? ~ pc" test, which uses:

```scala
val left = pa.? ~ (pb.? ~ pc)
val right = pa.? ~ pb.? ~ pc
```

In terms of error reporting these two parsers should report (basically) the same errors, which is tested here with the empty string input.

Now, if trying to parse "a", we should get that "b" or "c" are possible at offset=1 from both parsers above, correct?

Well, while that's the case with `left`:

```scala
    val usingLeft = left.parse("a").void
    assertEquals(
      usingLeft.leftMap(_.expected),
      Left(
        NonEmptyList.of(
          Parser.Expectation.InRange(1, 'b', 'c')
        )
      )
    )
```

it is not with `right`:

```scala
    val usingRight = right.parse("a").void
    assertEquals(
      usingRight.leftMap(_.expected),
      Left(
        NonEmptyList.of(
          Parser.Expectation.InRange(1, 'b', 'c')
        )
      )
    )
```

```scala
2996:    val usingRight = right.parse("a").void
2997:    assertEquals(
2998:      usingRight.leftMap(_.expected),
values are not the same
=> Obtained
Left(
  value = NonEmptyList(
    head = InRange(
      offset = 1,
      lower = 'c',
      upper = 'c'
    ),
    tail = Nil
  )
)
=> Diff (- obtained, + expected)
       offset = 1,
-      lower = 'c',
+      lower = 'b',
       upper = 'c'
```

Thoughts?

johnynek:
well, that's certainly a good test to add...

And kudos to you for finding these issues...

I'll keep stabbing at this (and try to find a better law that would have found this without your diligent work).

johnynek:
Okay, actually, this whole approach has a serious issue:

The current code conflates OneOf parsing with OneOf error reporting. In parsing, you want to be as eager as possible especially in the case that you can eventually succeed. So, for instance, even if you have:

```
(a + b) * (c + d) = a * c + a * d + b * c + b * d
```
the left hand side first checks a or b THEN it checks c or d. Thus, you in the worst case check a + b + c + d if you parse bd: that's four checks. In the right hand side, if only bd succeeds, you could check a + c, then a + d, then b + c, then b + d for 8 checks.

So, I think actually this whole approach is a bit broken: we don't want to deoptimize parsing to report better errors I think. I think if we aren't careful, we are indeed doing that.

So, I want to make sure to run the benchmarks and check for any regressions.

Also, I did notice the tests taking longer, which should have been a warning to me.

carueda:
@johnynek 

Fully agree, that's why I've suggested some sort of opt-in mechanism for enriched error reporting; otherwise, provide best performance possible.

Again, I really very much appreciate your efforts, and I'm happy to keep providing feedback as much as I can, but I don't really have any pressing need for this feature: I'm using the `?:` operator/extension approach pretty satisfactorily. cats-parse is great!

johnynek:
@carueda okay, I'm a stubborn guy...

I think we can actually solve it for the narrow case we are interested in, which I think is worth it since I think it may be the most common use case of `Parser0`.

I guess a similar issue also exists with `a.rep0 ~ b == (a.rep ~ b) | b` but I'm not going to address that one in this PR.

I think this may be mergable. What do you think?

johnynek:
Benchmarks:
```
main:

[info] Benchmark                          (test)  Mode  Cnt      Score      Error  Units
[info] BarBench.catsParseParse               N/A  avgt    5     ≈ 10⁻⁴             ms/op
[info] Bla25Bench.catsParseParse             N/A  avgt    5     23.394 ±    0.091  ms/op
[info] CountriesBench.catsParseParse         N/A  avgt    5      7.415 ±    0.245  ms/op
[info] Qux2Bench.catsParseParse              N/A  avgt    5      7.024 ±    0.050  ms/op
[info] StringInBenchmarks.linearMatchIn      foo  avgt    5     78.992 ±    0.036  ns/op
[info] StringInBenchmarks.linearMatchIn    broad  avgt    5  93871.853 ± 1517.660  ns/op
[info] StringInBenchmarks.oneOfParse         foo  avgt    5     97.041 ±   12.488  ns/op
[info] StringInBenchmarks.oneOfParse       broad  avgt    5   1018.093 ±    3.529  ns/op
[info] StringInBenchmarks.radixMatchIn       foo  avgt    5     59.462 ±    0.246  ns/op
[info] StringInBenchmarks.radixMatchIn     broad  avgt    5    748.578 ±   10.112  ns/op
[info] StringInBenchmarks.stringInSParse     foo  avgt    5     72.742 ±    0.279  ns/op
[info] StringInBenchmarks.stringInSParse   broad  avgt    5   1490.265 ±    5.512  ns/op
[info] StringInBenchmarks.stringInVParse     foo  avgt    5     97.500 ±    4.395  ns/op
[info] StringInBenchmarks.stringInVParse   broad  avgt    5   1018.446 ±   14.278  ns/op
[info] Ugh10kBench.catsParseParse            N/A  avgt    5     59.917 ±    0.172  ms/op
[success] Total time: 1997 s (33:17), completed Mar 27, 2022 10:01:22 AM

a.? ~ b optimization:
[info] Benchmark                          (test)  Mode  Cnt      Score     Error  Units
[info] BarBench.catsParseParse               N/A  avgt    5     ≈ 10⁻⁴            ms/op
[info] Bla25Bench.catsParseParse             N/A  avgt    5     23.278 ±   0.120  ms/op
[info] CountriesBench.catsParseParse         N/A  avgt    5      7.424 ±   0.020  ms/op
[info] Qux2Bench.catsParseParse              N/A  avgt    5      6.910 ±   0.047  ms/op
[info] StringInBenchmarks.linearMatchIn      foo  avgt    5     78.555 ±   0.307  ns/op
[info] StringInBenchmarks.linearMatchIn    broad  avgt    5  93210.906 ± 149.028  ns/op
[info] StringInBenchmarks.oneOfParse         foo  avgt    5    607.254 ±   1.042  ns/op
[info] StringInBenchmarks.oneOfParse       broad  avgt    5   1603.108 ±   2.166  ns/op
[info] StringInBenchmarks.radixMatchIn       foo  avgt    5     59.488 ±   0.040  ns/op
[info] StringInBenchmarks.radixMatchIn     broad  avgt    5    811.793 ±   1.801  ns/op
[info] StringInBenchmarks.stringInSParse     foo  avgt    5    542.707 ±   1.421  ns/op
[info] StringInBenchmarks.stringInSParse   broad  avgt    5   1548.265 ±   1.656  ns/op
[info] StringInBenchmarks.stringInVParse     foo  avgt    5    550.702 ±   0.858  ns/op
[info] StringInBenchmarks.stringInVParse   broad  avgt    5   1605.281 ±   1.770  ns/op
[info] Ugh10kBench.catsParseParse            N/A  avgt    5     52.888 ±   0.196  ms/op
[success] Total time: 1518 s (25:18), completed Mar 27, 2022 10:30:22 AM
```

So, it looks like this PR has somewhat helped the json parsing benchmark but done considerable harm to `stringIn` which I didn't expect at all. It may be possible to remedy that, but if not, we should probably not merge.

johnynek:
oh, I bet I know the mechanism: the `parseAll` change. Since that now has a lazy val, we defer computing the `~ Parser.end` until we are actually parsing something, thus, we are including the parser optimization in the first parse, which we were not before.

Nothing else should have changed. So, if we could revert that part, maybe this becomes an across the board win.

johnynek:
okay, after reverting the parseAll change the benchmarks look slightly positive:
```
[info] Benchmark                          (test)  Mode  Cnt      Score    Error  Units
[info] BarBench.catsParseParse               N/A  avgt    5     ≈ 10⁻⁴           ms/op
[info] Bla25Bench.catsParseParse             N/A  avgt    5     23.282 ±  0.108  ms/op
[info] CountriesBench.catsParseParse         N/A  avgt    5      7.337 ±  0.017  ms/op
[info] Qux2Bench.catsParseParse              N/A  avgt    5      6.821 ±  0.012  ms/op
[info] StringInBenchmarks.linearMatchIn      foo  avgt    5     78.425 ±  0.104  ns/op
[info] StringInBenchmarks.linearMatchIn    broad  avgt    5  93151.523 ± 87.992  ns/op
[info] StringInBenchmarks.oneOfParse         foo  avgt    5     98.287 ±  4.869  ns/op
[info] StringInBenchmarks.oneOfParse       broad  avgt    5   1044.964 ± 55.673  ns/op
[info] StringInBenchmarks.radixMatchIn       foo  avgt    5     59.198 ±  3.007  ns/op
[info] StringInBenchmarks.radixMatchIn     broad  avgt    5    743.086 ± 13.285  ns/op
[info] StringInBenchmarks.stringInSParse     foo  avgt    5     71.911 ±  0.303  ns/op
[info] StringInBenchmarks.stringInSParse   broad  avgt    5   1009.045 ±  0.486  ns/op
[info] StringInBenchmarks.stringInVParse     foo  avgt    5     97.745 ±  3.570  ns/op
[info] StringInBenchmarks.stringInVParse   broad  avgt    5   1024.573 ±  0.840  ns/op
[info] Ugh10kBench.catsParseParse            N/A  avgt    5     53.545 ±  0.127  ms/op
[success] Total time: 2580 s (43:00), completed Mar 27, 2022 11:48:01 AM
```

johnynek:
@carueda if you give the +1 I'll merge this.

I'm looking to publish in the next 7 days.

carueda:
ok, pulling and testing shortly ...

carueda:
@johnynek 👍🏻   

Thanks again!

PR #396, opened 2022-03-27
Improve error reporting for rep0, repAs0

johnynek:
give rep0/repAs0 the same treatment as #389


johnynek:
cc @carueda

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -2952,22 +2959,6 @@ object Parser {
       }
     }
 
-    case class Rep0[A, B](p1: Parser[A], maxMinusOne: Int, acc: Accumulator0[A, B])
```
johnynek, line 2955: in the presence of the `a.? ~ b` optimization, if we want consistency, it is better to just remove `Rep0` entirely and represent it in the same way so we don't have any chance of breaking the symmetry there.

regadas (approved):

PR #410, opened 2022-04-28
Implement Show[Error]

MasseGuillaume:


core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -821,11 +858,60 @@ object Parser {
 
   /** Represents where a failure occurred and all the expectations that were broken
     */
-  final case class Error(failedAtOffset: Int, expected: NonEmptyList[Expectation]) {
+  final case class Error(input: String, failedAtOffset: Int, expected: NonEmptyList[Expectation]) {
```
johnynek, line 861: this is binary and source incompatible and we have been very reluctant to break people.

I might suggest we add a new case class:

```scala
final case class InputAndError(input: String, error: Error)
```
add `def withInput(input: String): InputAndError` to Error and then make the Show instance on that.

How does that sound?

It is probably a design flaw we didn't include the input with the Error, but I really don't want to break compatibility.

Maybe we can find a source and binary compatible way to add the input (for instance, make Error not a case class, but manually add back all the methods that are generated).

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -821,11 +858,60 @@ object Parser {
...
 
+  object Error {
+    implicit val catsShowError: Show[Error] =
+      new Show[Error] {
+        def show(error: Error): String = {
+          val locationMap = new LocationMap(error.input)
+          val errorMsg = error.expected.map(e => s"* ${e.show}").toList.mkString("\n")
+
+          locationMap.toCaret(error.failedAtOffset).fold(errorMsg) { caret =>
+            val lines = error.input.split('\n')
```
johnynek, line 874: the `LocationMap` has the lines and you can get it from there. We shouldn't split it again.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -821,11 +858,60 @@ object Parser {
...
     def offsets: NonEmptyList[Int] =
       expected.map(_.offset).distinct
   }
 
+  object Error {
+    implicit val catsShowError: Show[Error] =
+      new Show[Error] {
+        def show(error: Error): String = {
+          val locationMap = new LocationMap(error.input)
+          val errorMsg = error.expected.map(e => s"* ${e.show}").toList.mkString("\n")
```
johnynek, line 871: this should be a `def` or `lazy val` otherwise we are evaluating both branches all the time.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -821,11 +858,60 @@ object Parser {
...
+
+          locationMap.toCaret(error.failedAtOffset).fold(errorMsg) { caret =>
```
johnynek, line 873: let's use so it is explicit that we are not evaluating both branches.
```scala
locationMap.toCaret(error.failedAtOffset) match {
  case None => error.expected.map(e => s"* ${e.show}").toList.mkString("\n")
  case Some(caret) =>
     ...
}
```

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -821,11 +858,60 @@ object Parser {
...
+
+            val beforeElipsis: Option[String] =
+              if (start <= 0) None
+              else Some(elipsis)
+
+            val beforeContext: Option[String] =
+              Some(lines.slice(start, caret.line).mkString("\n")).filter(_.nonEmpty)
+
+            val line: Option[String] =
+              if (caret.line >= 0 && caret.line < lines.length) Some(lines(caret.line))
```
johnynek, line 891: this has to be true by construction of the `Caret` I believe?

core/shared/src/test/scala/cats/parse/ErrorShowTest.scala
```diff
@@ -0,0 +1,182 @@
...
+  error(
+    lxOk.rep(2),
+    """|ko
+       |l2""".stripMargin,
+    """|ko
+       |^
+       |* one of: {"ok"}
+       |* is: l
+       |l2""".stripMargin
+  )
```
johnynek, line 181: can we also add a scalacheck that does something simple like:
```scala
forAll { (p: Parser[Any], in: String) =>
  p.parseAll(in) match {
    case Right(_) => ()
    case Left(err) =>
      // make sure we exercise all the paths in show, and don't throw.
      assert(err.show ne null)
  }
}
```

core/shared/src/test/scala/cats/parse/ErrorShowTest.scala
```diff
@@ -0,0 +1,182 @@
...
+
+  def error(parser: Parser0[Any], input: String, expected: String)(implicit
+      loc: munit.Location
+  ): Unit = {
+    test(input) {
+      parser
+        .parseAll(input)
+        .fold(
+          e =>
+            assert(
```
johnynek, line 18: Let's use `assertEquals` which does better diffing I think.

johnynek (commented):
Thanks so much for this! A number of people have asked.

I just want to find a way to avoid breaking people.

core/shared/src/test/scala/cats/parse/ErrorShowTest.scala
```diff
@@ -0,0 +1,182 @@
...
+  )
+
+  // Length
+  error(
+    ok | length(2),
+    "aaaaaaaaaaaaaaaaaaa",
+    """|aaaaaaaaaaaaaaaaaaa
+       |  ^
+       |* end of string, length = 19""".stripMargin
+  ) // this does repport the correct error IMO
```
MasseGuillaume, line 84: @johnynek do you know how can I trigger this Error?

johnynek: did you mean "this does *not* report"?

I would say something like `expected end of string but 19 characters remaining`.

This issue has a discussion about `|` hiding error messages:

https://github.com/typelevel/cats-parse/issues/382

core/shared/src/test/scala/cats/parse/ErrorShowTest.scala
MasseGuillaume: Can you give me more details, I'm not sure I understand.

Is the idea here to generate arbitrary parsers and arbitrary output to have full branch coverage?

How do I construct an `Arbitrary[Parser[Any]]` the search space must be huge.

johnynek: Yes, the idea is to ensure that all branches are exercised and we don't have some cases where even basic things fail.

https://github.com/typelevel/cats-parse/blob/e31eb1172ecb268f782c7f4c2845b905bcd1f849/core/shared/src/test/scala/cats/parse/ParserTest.scala#L602

there is how you generate an arbitrary parser. The space is large, but it is already implemented (and how I think we have had so few bugs).

core/shared/src/main/scala/cats/parse/Parser.scala
MasseGuillaume, line 861: would it be ok if I override toString with the show instance?

MasseGuillaume: I Implemented the second option where I coded manually what the Scala compiler would output for a case class

core/shared/src/test/scala/cats/parse/ErrorShowTest.scala
MasseGuillaume: ah I think I figured it out, length is only trigger when the output is smaller

ex: `length(2).parseAll("a")`

core/shared/src/test/scala/cats/parse/ErrorShowTest.scala
MasseGuillaume: Hum, I get:

```
[error] /home/gui/foos/cats-parse/core/shared/src/test/scala/cats/parse/ErrorShowTest.scala:180:12: diverging implicit expansion for type org.scalacheck.Arbitrary[A]
[error] starting with method arbFunction0 in object Arbitrary
[error]     forAll { (p: Parser[Any], in: String) => 
[error]            ^
[error] one error found
```

johnynek: I think it is the `Any` there. Try `forAll { (p: Parser[Unit], in: String) =>`

johnynek: assuming you have imported the method I linked to.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -821,9 +858,119 @@ object Parser {
...
   /** Represents where a failure occurred and all the expectations that were broken
     */
-  final case class Error(failedAtOffset: Int, expected: NonEmptyList[Expectation]) {
+  final class ErrorWithInput(
+      val input: String,
+      override val failedAtOffset: Int,
+      override val expected: NonEmptyList[Expectation]
+  ) extends Error(failedAtOffset, expected) {}
+
+  class Error(val failedAtOffset: Int, val expected: NonEmptyList[Expectation])
```
johnynek, line 867: if we go this route, I'd rather a single class:

```scala
class Error(val input: Option[String], val failedAtOffset: Int, val expected: NonEmptyList[Expectation]) {
  def this(failedAtOffset: Int, expected: NonEmptyList[Expectation]) =
    this(None, failedAtOffset, expected)
}
```
then we can also include the `input: Option[String]` in the `def equals`.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -821,9 +858,119 @@ object Parser {
...
+
+      } else false
+    }
+  }
+
+  object Error extends Serializable {
+    def apply(failedAtOffset: Int, expected: NonEmptyList[Expectation]): Error =
+      new Error(failedAtOffset, expected)
+
+    def unapply(error: Error): Option[(Int, NonEmptyList[Expectation])] =
```
johnynek, line 920: if you make the return type `Some` I think the compiler treats this differently, but that might be a mima error.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -742,6 +745,40 @@ object Parser {
...
+
+          case InRange(_, lower: Char, upper: Char) =>
+            if (lower != upper) s"in range: ['$lower', '$upper']"
+            else s"is: '$lower'"
+
+          case StartOfString(_) =>
+            "start of string"
+
+          case EndOfString(_, length) =>
+            s"end of string, length = $length"
```
johnynek, line 763: I think `end of string, but found $length chars remaining` or something might be better. Length is unclear as an error message. Or, you can simply write `end of string` if you are already pointing to the location is may be pretty obvious how much is left.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -742,6 +745,40 @@ object Parser {
...
+
+          case Length(_, expected, actual) =>
+            s"length: expected = $expected, actual = $actual"
+
+          case ExpectedFailureAt(_, matched) =>
+            s"failure at $matched"
```
johnynek, line 769: I think this is unclear, I think it should be `expected failure but matched $matched` or something. This expectation means we *didn't* fail when we thought we would. Note, it is confusing that this and `fail` and `fail: ` all start the same but are rather different.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -821,9 +858,131 @@ object Parser {
...
+                    Some(line),
+                    Some((1 to caret.col).map(_ => " ").mkString("") + "^"),
+                    Some(errorMsg),
+                    afterContext,
+                    afterElipsis
+                  ).flatten.mkString("\n")
+                }
+              }
+            }
+            case None => "no input"
```
johnynek, line 982: can we just use total offsets here and not print nothing? We can still show the `Expectations` right?

build.sbt
```diff
@@ -149,6 +149,14 @@ lazy val core = crossProject(JSPlatform, JVMPlatform)
           ProblemFilters.exclude[IncompatibleResultTypeProblem]("cats.parse.Parser#State.error")
         )
       else Nil
+    } ++ {
+      if (tlIsScala3.value)
+        List(
```
johnynek, line 154: can we merge this with the above list?

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -750,32 +750,32 @@ object Parser {
         private val dq = "\""
         def show(expectation: Expectation): String = expectation match {
           case OneOfStr(_, strs: List[String]) =>
-            "one of: " + strs.map(s => dq + s + dq).mkString("{", ", ", "}")
+            "must match one of the strings: " + strs.map(s => dq + s + dq).mkString("{", ", ", "}")
```
johnynek, line 753: nit: `strs.iterator.map(..).mkString` will be faster and avoid allocating an intermediate list.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -921,23 +924,33 @@ object Parser {
...
+    def unapply(error: Error): Option[(Int, NonEmptyList[Expectation])] =
       Some((error.failedAtOffset, error.expected))
 
     implicit val catsShowError: Show[Error] =
       new Show[Error] {
         def show(error: Error): String = {
+          val nl = "\n"
+
+          def errorMsg = {
+            val expectations = error.expected.map(e => s"* ${e.show}").toList.mkString(nl)
```
johnynek, line 945: nit, `error.expected.toList.iterator.map(...).mkString` will avoid reallocating the whole list.

core/shared/src/main/scala/cats/parse/Parser.scala
```diff
@@ -921,23 +924,33 @@ object Parser {
...
+
+            s"""|expectation${if (error.expected.length > 1) "s" else ""}:
```
johnynek, line 947: `error.expected.lengthCompare(1) > 0`, `.length` is O(N).

johnynek, line 947: `error.expected.lengthCompare(1) > 0`, `.length` is O(N).

core/shared/src/test/scala/cats/parse/ErrorShowTest.scala
```diff
@@ -130,8 +139,9 @@ class ErrorShowTest extends munit.ScalaCheckSuite {
        |l4
        |ko
        |^
-       |* one of: {"ok"}
-       |* is: 'l'
+       |expectations:
+       |* must match one of the strings: {"ok"}
```
johnynek, line 143: this is kind of strange right? Maybe we should say `must match string "ok"` if there is only one item in that list?

johnynek:
sorry I lost track of this.

This needs a format (`prePR` I think will do it). Then I think we can merge.

I will publish a new version afterwards.

johnynek:
Thanks for your work on this PR! I think many will appreciate it.

PR #420, opened 2022-06-11
Only publish site on release

regadas:
Fixing the current `main` build issue introduced with https://github.com/typelevel/cats-parse/runs/6804145548

Including a Java step here as well since parboiled, dependency from paradox, is java 11 only now.

We had this before, but we can change CI to use Java 11 by default since we are setting the targets correctly.


johnynek (approved):
Thanks!

Issue #433, opened 2022-07-27
add json string parsing


johnynek:
String parsing with escapes is a bit painful.

We don't want to have a json ast or depend on a json library to get one, but if you embed json in the language you are parsing, it can be convenient to at least provide json string parsing since we already have json number parsing, and the rest of json parsing is really only a few lines of code.

PR #437, opened 2022-08-07
Test json number parsing against Jawn

johnynek:
The idea here is to make sure we actually agree with a json number parsing implementation such as Jawn.

We should do the same for json string parsing when that is added #433


johnynek:
I want to copy some of this to solve #433 so merging without other approval. Since it is only test code and we can revert if needed.

PR #468, opened 2022-12-07
Add an example on how to implement a parser for interspersed input

daddykotex:
While looking at the documentation I found some things that I could be improved so here is a PR for that.

I was specifically looking to see how I could implement something to parse `"a,b,c" and extract `a`, `b` and `c`. I found that soft could be used for that and because I think it's a common pattern in parsers, I figured it could be useful to have an example in the documentation.

On top of that, I get feedback from experts to tell me if that solution is bad and what would be a better alternative, win-win!

----

One annoying thing about the `docs/index.md` is that yes, it's compiled but it's not the actual output of mdoc and so the result in there are hard coded. I found it annoying to go into `site/target/mdoc/index.md` and retrieve the output and format it to fit with the rest of the document. Is there a better way?

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -367,6 +367,31 @@ p3.parse("The Wind Has Risen")
 
 So when the _right side_ returns an epsilon failure the `soft` method allows us to rewind parsed input and try to proceed it's parsing with next parsers (without changing the parser itself!).
 
+Another common use case for `soft` is implementing a parser to find values interspersed with a separator. For example, if you want to extract `'a'`, `'b'` and `'c'` from `"a,b,c"`, you can use `soft`.
+
+Naively, one may try the following:
+
+```scala mdoc
+val p4 = alpha
+val naiveInterspersed = (p4 <* pchar(',')).rep
+naiveInterspersed.parse("a,b,c")
+// res4 = Left(Error(5,NonEmptyList(InRange(offset = 5, lower = ',', upper = ','),List())))
+```
+
+Basically, it's looking for the trailing comma:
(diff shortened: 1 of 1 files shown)


armanbilge:
> t's compiled but it's not the actual output of mdoc and so the result in there are hard coded ... Is there a better way?

mdoc already renders expressions automatically (but not statements like `val` declarations).

In fact I think in some places there is duplicate output.
https://typelevel.org/cats-parse/

docs/index.md
```diff
@@ -359,16 +357,41 @@ But this problem might be resolved with `soft` method inside the first parser si
...
+Basically, it's looking for the trailing comma:
+
+```scala mdoc
+naiveInterspersed.parse("a,b,c,")
+// res5 = Right(("", NonEmptyList('a', List('b', 'c'))))
+```
+
+But you can use, `soft` along with `|` to implement that:
+```scala mdoc
+val interspersed = ((p4.soft <* pchar(',')) | p4).rep
```
johnynek, line 390: this example would also parse `abc` which maybe what you want but might be a bit confusing for newcomers.

here is an example of how I parse lists in bosatsu:
https://github.com/johnynek/bosatsu/blob/c58b55785b6ac72e59b98afe3149a042d623a902/core/src/main/scala/org/bykn/bosatsu/Parser.scala#L302

johnynek (commented):
thanks for sending the PR.

I think we should add some notes to the example you added because I think it accepts more than most people would expect with a list example (making the , fully optional).

PR #524, opened 2023-08-09
Update to sbt-typelevel 0.5.0-RC11

armanbilge:
Updates to sbt-typelevel 0.5.0 series. Staying in RC phase while upgrading Typelevel projects to iron out any issues, but is otherwise stable.

For how much longer is the plan to keep building for Scala 2.11? I was able to make things work but it was mildly annoying.


johnynek (approved):
Thanks!

I'd like to keep 2.11 as long as I'm stuck on it at work so I can still use this project at work.

PR #555, opened 2023-11-06
Bump base version to 1.0

armanbilge:
Since cats-parse is ready for a v1.0.0 release 🎉 

CI will continue enforcing bincompat with the entire 0.3.x series. Thus, the v1.0.0 release will simply be a "rebranding" and is not a breaking release.

cc @milessabin


johnynek (approved):
Thank you!

PR #576, opened 2024-03-15
Use tlCommandAlias for command aliases

regadas:
`tlReplaceCommandAlias` is deprecated. Use `tlCommandAlias` instead.


johnynek (approved):
Thank you

Issue #435, opened 2022-08-03
Should `Parser` have a typeclass?


morgen-peschke:
One of the things I've noticed while using the library is using `Parser`s tends to follow this pattern (at least for me):
```
Foo.parser.parseAll(raw)
```

This has a couple of minor pain points (notably, "what did I name the parser?"), which should be straightforward to fix. Would you be interested in a PR that adds the needed scaffolding to allow this sort of thing instead:
```
import cats.parser.syntax._

raw.parseAll[Foo]
```

I'm planning on implementing this in one of my weekend projects anyway, so if you're interested I'll implement it as a PR instead of just something local to my project.

johnynek:
This is an interesting question. I personally think typeclasses work best when there is only one reasonable implementation. I don't think that is usually the case for parsers (consider `Parser[Int]` might not be very uncommon).

That said, you could imagine something similar to the scalacheck `Arbitrary/Gen` design, where `Gen` is not a typeclass, but `Arbitrary` is.

so, you could imagine something like:

```scala
trait DefaultParser0[A] {
  def parser0: Parser0[A]
}

object syntax {
  final case class ParserOps(private val str: String) extends AnyVal {
    def parseAll[A](implicit dp: DefaultParser0[A]): Either[Parser.Error, A] = dp.parser0.parseAll(str)
  }
}
```

Is that the kind of thing you are thinking about? I think something like this may work. Maybe we can hash out a rough design in this issue before we commit too much. What do you think?

morgen-peschke:
Yep, that's pretty much what I was thinking.

I agree that default parsers for most things don't really make sense, so we wouldn't need much in the way of default instances. Something like this should be enough:
```
trait Parseable[A] {
  def parser0: Parser0[A]
}
object Parseable {
  def apply[A](implicit P: Parseable[A]): P.type = P

  def instance[A](p: Parser0[A]): Parseable[A] = new Impl[A](p)

  private final class Impl[A](override val parser0: Parser0[A])
      extends Parseable[A]
      with Serializable
}
```
`Parseable.instance` and `Impl` are because the chance that ScalaTest will close over a `Parseable` instance is pretty good, and it chokes if it grabs anything that doesn't extend `Serializable`. 

For the syntax, it might make sense to include `parse` as well as `parseAll`:
```
object syntax {
  implicit final class ParseableOps(private val raw: String) extends AnyVal {
    def parse[A: Parseable]: Either[Parser.Error, (String, A)] =
[...]

PR #577, opened 2024-03-28
Add DefaultParser and DefaultParser0 typeclasses

morgen-peschke:
These provide a way to bless canonical `Parser`/`Parser0` instances and easily call them.

Implements #435

### DIFF
--- a/core/shared/src/main/scala/cats/parse/extra/DefaultParser.scala
+++ b/core/shared/src/main/scala/cats/parse/extra/DefaultParser.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024 Typelevel
+ * Copyright (c) 2021 Typelevel
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy of
  * this software and associated documentation files (the "Software"), to deal in
--- a/core/shared/src/main/scala/cats/parse/extra/DefaultParser.scala
+++ b/core/shared/src/main/scala/cats/parse/extra/DefaultParser.scala
@@ -26,7 +26,7 @@ import cats.parse.Parser
 /** Typeclass for "has a Parser"
   *
   * This is primarily provided to help keep track of `Parser` instances, and as such the omission of
-  * `cats.Functor` instances is intentional.
+  * a `cats.Functor` instance is intentional.
   * @tparam A
   */
 trait DefaultParser[+A] {
--- a/core/shared/src/main/scala/cats/parse/extra/DefaultParser.scala
+++ b/core/shared/src/main/scala/cats/parse/extra/DefaultParser.scala
@@ -31,6 +31,18 @@ import cats.parse.Parser
   */
 trait DefaultParser[+A] {
   def parser: Parser[A]
+
+  /** Pass through to equivalent method on `Parser`
+    * @see
+    *   [[Parser.parse]]
+    */
+  def parse(string: String): Either[Parser.Error, (String, A)] = parser.parse(string)
+
+  /** Pass through to equivalent method on `Parser`
+    * @see
+    *   [[Parser.parseAll]]
+    */
+  def parseAll(string: String): Either[Parser.Error, A] = parser.parseAll(string)
 }
 object DefaultParser {
   def apply[A](implicit P: DefaultParser[A]): P.type = P
--- a/core/shared/src/main/scala/cats/parse/extra/DefaultParser.scala
+++ b/core/shared/src/main/scala/cats/parse/extra/DefaultParser.scala
@@ -23,13 +23,12 @@ package cats.parse.extra
 
 import cats.parse.Parser
 
-/**
- * Typeclass for "has a Parser"
- *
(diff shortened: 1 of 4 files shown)


core/shared/src/main/scala/cats/parse/extra/DefaultParser0.scala
```diff
@@ -0,0 +1,36 @@
...
+      def parse[A: DefaultParser0]: Either[Parser.Error, (String, A)] =
+        DefaultParser0[A].parser0.parse(raw)
+
+      def parseAll[A: DefaultParser0]: Either[Parser.Error, A] =
+        DefaultParser0[A].parser0.parseAll(raw)
+    }
+  }
+
+  implicit def defaultParserIsDefaultParser0[A: DefaultParser]: DefaultParser0[A] =
+    DefaultParser0.instance(DefaultParser[A].parser)
```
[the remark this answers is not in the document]
morgen-peschke: Not particularly, I'll occasionally include the object for methods that are kind of generic.

I don't have strong feelings about it, and am happy to remove it if you prefer

morgen-peschke:
> I also wonder a bit about this typeclass being lawless.
> 
> It could be a lawful typeclass if paired with a function that converts A to String and the law is that we can always parse and get back the original.
> 
> What do you think about the lawlessness?

I don't have any strong opinions about it being lawless.

I have the mild inclination that creating laws for `DefaultParser0`/`DefaultParser0` would imply the existence of `Coparser`/`Coparser0`, but I have no idea if that's worth representing in the code as something other than `A => String`.

It'd probably make testing parsers over the set of valid inputs easier, so it might be worth doing for that reason alone.

regadas (commented):
lgtm! 👍 but I'll let @johnynek do the follow up!

regadas (approved):
lgtm! 👍 but I'll let @johnynek do the follow up!

core/shared/src/main/scala/cats/parse/extra/DefaultParser.scala
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+ * FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+ * COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package cats.parse.extra
```
johnynek, line 22: I think we should put this in the main package. There are very few names in cats.parse already.

core/shared/src/main/scala/cats/parse/extra/DefaultParser0.scala
```diff
@@ -0,0 +1,56 @@
...
+import cats.parse.{Parser, Parser0}
+
+/** Typeclass for "has a Parser0"
+  *
+  * This is primarily provided to help keep track of `Parser0` instances, and as such the omission
+  * of `cats.Functor` instances is intentional.
+  * @tparam A
+  */
+trait DefaultParser0[+A] {
+  def parser0: Parser0[A]
```
johnynek, line 33: Can we add plumbing through parse and parseAll here so you can do: `DefaultParser[Foo].parseAll(s)`

johnynek (changes requested):
Sorry for the latency on my part.

Two minor requests then I'll merge.

Thanks for sending this PR.

core/shared/src/main/scala/cats/parse/extra/DefaultParser.scala
morgen-peschke: Will do 👍🏻

core/shared/src/main/scala/cats/parse/extra/DefaultParser0.scala
morgen-peschke: Easy enough 👍🏻

johnynek: I don't see this change yet. Did you reconsider this?

morgen-peschke: Looks like it's up, my guess is it took a couple minutes to propagate to the UI

regadas: this was a good idea 🙌

johnynek (approved):

morgen-peschke:
Thanks for everyone's help :)

PR #603, opened 2024-07-31
Reduce tutorial mdoc noise

valencik:
Current there is duplication / noise in the mdoc output from hardcoded result comments:

Screenshot 2024-07-30 at 10 08 15 PM

This PR removes these comments as mdoc computes and displays them for us.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -41,11 +41,10 @@ import cats.parse.Parser
 val p: Parser[Char] = Parser.anyChar
 
 p.parse("t")
-// res0: Either[Error, Tuple2[String, Char]] = Right((,t))
+
 p.parse("")
-// res1: Either[Error, Tuple2[String, Char]] = Left(Error(0,NonEmptyList(InRange(0,,))))
+
 p.parse("two")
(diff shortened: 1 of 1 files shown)


johnynek:
I think without seeing the output, those examples don't really work. There are now just a lot of calls to `.parse`. What are they returning?

It isn't clear.

If we do this, we should add some asserts or something so people can understand if the parse succeeds or fails.

valencik:
Hey Oscar, that's for taking a look at this :)

Just to be clear, the output will still appear on the site, mdoc executes the code and inserts the results with their types in the final output.
This PR is just removing the additional "manual" result line comments, which are not type checked by mdoc, and add extra noise to the site output.

Here's a Before and After screenshot:

**Before**
cats-parse-before

**After**
cats-parse-after


This is also nice because mdoc is a bit more clear in how it prints some output.
Compare mdoc's `Right(value = ("", 't'))` to the prior manual annotation `Right((,t))`

johnynek:
got it, thanks for clarifying.

PR #624, opened 2024-11-14
Implicitly use Show[Expectation]

### LINKED ISSUE #607: Not possible to provide own Show[Expectation] without Show[Error]
Recently I've needed to provide my own `Show[Expectation]` but wanted to keep `Show[Error]` as is. It turns out it's not possible, as the latter doesn't take `Show[Expectation]` as a parameter. Example code (runnable via scala-cli):
```scala
//> using dep "org.typelevel::cats-parse:1.0.0"
//> using scala 2.13

import cats.parse.Parser
import cats.parse.Parser.{Error, Expectation}

import cats.Show
import cats.syntax.show._

val parseError: Error = Parser.string("something").parseAll("smth").swap.toOption.get

implicit val showExpectation: Show[Expectation] =
  new Show[Expectation] {
    override def show(exp: Expectation) = "myImpl"
  }

println(show"$parseError")

/* prints this (showExpectation is apparently ignored):

  smth
  ^
  expectation:
  * must match string: "something"

*/
```
It appears to be necessary to copy `Show[Error]` verbatim if one wants to customise how the expectations are rendered.

I think it wouldn't be necessary if this definition:
```scala
implicit val catsShowError: Show[Error] =
```
were changed to
```scala
implicit def catsShowError(implicit showExpectation: Show[Expectation]): Show[Error] =
```

To be honest I know my use case is very, very niche, but I thought I'd share this suggestion in case it appears useful. On the other hand, if the suggestion is wrong for some reason, I'll be happy to learn

johnynek:

close #607 

cc @DLakomy

### DIFF
--- a/core/shared/src/main/scala/cats/parse/Parser.scala
+++ b/core/shared/src/main/scala/cats/parse/Parser.scala
@@ -974,13 +974,13 @@ object Parser {
     def unapply(error: Error): Option[(Int, NonEmptyList[Expectation])] =
       Some((error.failedAtOffset, error.expected))
 
-    implicit val catsShowError: Show[Error] =
+    implicit def catsShowErrorGivenExpectation(implicit showExp: Show[Expectation]): Show[Error] =
       new Show[Error] {
         def show(error: Error): String = {
           val nl = "\n"
 
           def errorMsg = {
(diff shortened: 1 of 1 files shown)


regadas (approved):
looks great!

PR #643, opened 2025-01-24
Add Align instance

johnynek:


johnynek:
Align isn't in scala 2.11 so we have to drop support for that first. I guess since others are talking about dropping 2.12, dropping 2.11 seems like it should be fine.

regadas (approved):
lgtm! 

@johnynek I guess a release with 2.11 dropped needs to go out first right?

johnynek:
@regadas since the last release at the end of the year, there haven't been any changes except build dependencies:

https://github.com/typelevel/cats-parse/releases

I already removed 2.11 in a previous PR to get sbt-typelevel to work: https://github.com/typelevel/cats-parse/pull/644

So I think we can just merge this as is...

So far it would be the only user facing change (other than no more 2.11, but if you want 2.11, just use the last published version I guess).

regadas:
@johnynek sounds good to me!

PR #644, opened 2025-01-24
Drop scala 2.11

johnynek:


johnynek:
includes #642

johnynek:
includes also #618

